PR #7, opened 2017-03-30
Add contribution instructions to README

deltaidea:
Add "Getting Started", "Built With" sections to `README.md`, and a reference to the Unlicense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,60 @@
-
 # Hollowverse
 
-## UI guidelines
+Politics, religions, and ideas.
+
+# Contributing
+
+We're currently at the stage of heavy initial development. Any contributions are appreciated!
+Please submit an issue with you questions if you're not sure where to start.
+
+## Getting Started
+
+### Prerequisites
+
+* Install latest stable [Node.js](https://nodejs.org/en/)
+* Install dependencies: `npm install` in this directory
+
+### Front-end
+
+* Start webpack-dev-server: `npm run client/dev`
+
+This will provide a dev version of the front-end on `localhost:8080` and rebuild it on changes.
+
+### Back-end
+
+* Start the server via nodemon: `npm run server/dev`
+
+This will run the server straight from the source in `./server/src`. Nodemon will automatically refresh it on changes.
+
+## Built With
+
+* [TypeScript](https://www.typescriptlang.org/) - strongly typed language compiled to JavaScript
+* [React](https://facebook.github.io/react/) - front-end library for building UI
+* [Redux](http://redux.js.org/) - state container for structuring business logic in a predictable way
+* [Sass](http://sass-lang.com/) - superset of CSS for easier styling
+* [Firebase](https://firebase.google.com/) - tools and infrastructure for the server side, abstraction over the Google Cloud Platform
+* [Webpack](https://webpack.github.io/) - module bundler for concatenating and minifying JavaScript
+
+### UI guidelines
+
 * Layout should redraw as little as possible as data comes in
 * Follow the advice in [this article](https://goo.gl/1V7aJw)
 
-## Engineering approach
+### Engineering approach
+
 * Use strong static types to mitigate some of the need of unit testing
 * When a runtime bug is discovered, think about how to prevent it with stronger typing
 
-## React guidelines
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,17 +1,61 @@
-
 # Hollowverse
 
-## UI guidelines
+The religions and political views of the influentials
```
wholesomedev: I'm trying to think of a more lighthearted tagline, like maybe 

"Politics, religions, and ideas"

...or something. Any ideas?

README.md
```diff
@@ -1,17 +1,61 @@
...
+
+## Getting Started
+
+These instructions will get you a copy of the project up and running on your local machine for development and testing purposes. See deployment for notes on how to deploy the project on a live system.
```
wholesomedev: I don't think we should maintain instructions for deployment here. I don't see why anyone would want to deploy a clone of Hollowverse to a live environment.

README.md
```diff
@@ -1,17 +1,61 @@
...
+
+### Back-end
+
+* Start the server via nodemon: `npm run server/dev`
+
+This will run the serve straight from the source in `./server/src`. Nodemon will automatically refresh it on changes.
+
+## Deployment
+
+*TODO*
```
wholesomedev: Yeah, again, I'm not sure we should provide these instructions here.

README.md
```diff
@@ -1,17 +1,61 @@
...
+* [React]() - front-end library for building UI
+* [Redux]() - state container for structuring business logic in a predictable way
+* [Sass]() - superset of CSS for easier styling
+* [Firebase]() - tools and infrastructure for the server side, abstraction over the Google Cloud Platform
+* [Webpack]() - module bundler for concatenating and minifying JavaScript
+
+## Contributing
+
+We're currently at the stage of heavy initial development. Any contributions are appreciated!
+Please submit an issue with you questions if you're not sure where to start.
```
wholesomedev: "your* questions"

package.json
```diff
@@ -4,10 +4,10 @@
   "description": "Hollowverse.com",
   "main": "index.js",
   "scripts": {
-    "client/dev": "node_modules/.bin/webpack-dev-server --historyApiFallback true --content-base client/src --config client/webpack.dev.js",
-    "client/build": "node_modules/.bin/webpack --optimize-minimize --define process.env.NODE_ENV=\"'production'\" --config client/webpack.dev.js",
-    "zserver/dev": "node_modules/.bin/webpack --config server/webpack.dev.js",
-    "server/dev": "node_modules/.bin/nodemon --exec node_modules/.bin/ts-node -- server/src/index.tsx"
+    "client/dev": "webpack-dev-server --historyApiFallback true --content-base client/src --config client/webpack.dev.js",
```
wholesomedev, line 7: Isn't it better to have `node_modules/.bin/` in there to make sure we're always pointing to the local installation of the dependency rather than the global one in order to avoid potential inconsistencies that could result form different global versions?

README.md
deltaidea: _We_ certainly do want. This readme is primarily for ourselves. I intended to add instructions on how to build everything in production mode, and not the deployment process itself. This is purely so I don't forget how to build when I occasionally want to check that it works in prod.

The title is misleading though, you're right.

package.json
deltaidea: npm always prefers local packages to global. You can easily check that this is indeed true:

- `npm i -g nodemon@1.9.0` - we use 1.11.0
- add `"check": "nodemon --version"` to scripts
- `npm run check` - outputs 1.11.0, which is our local dependency

Don't forget to remove the unneeded global package.

README.md
wholesomedev: Ok

package.json
wholesomedev: Alright

README.md
wholesomedev: So should we change this before we merge the PR?

deltaidea:
Oops, accidentally closed by deleting `develop`.

deltaidea:
Made changes according to your comments and squashed.
Anything else here?

wholesomedev (approved):
LGTM

PR #16, opened 2017-04-01
Rearrange files

wholesomedev:
This PR puts files inside directories and abandons the flat file structure. I think the flat file structure was getting a bit unwieldy.


deltaidea (changes requested):
👍 Good idea.

The only comment I have is that tucking away `index.tsx` into `main/` is going into the opposite extreme. It's not really related to the `app.tsx` so closely either.

What do you think about keeping it in the `client/src/`? And renaming `main/` to `app/` since that's what's inside at that point.

wholesomedev:
Alright, I moved `index.ts` outside of `main` and renamed `main` to `app`

deltaidea (approved):

PR #19, opened 2017-04-03
Ignore `client/dist` and `server/dist`

deltaidea:
Keeping compiled files in the repo is not a great idea.

This is a duplicate of #8 but into `master`. We got rid of `develop` before this got merged further.


wholesomedev (approved):
LGTM

PR #22, opened 2017-04-06
Deployment

wholesomedev:
I have initted a Firebase project at the root of our repo. Basically, I followed the Firebase hosting instructions (https://firebase.google.com/docs/hosting/) and that asked me to install the Firebase CLI (https://firebase.google.com/docs/cli/) and init a project. I did that and here's the structure that they gave us:

* `/public` is gonna contain all the static assets (JS, CSS, HTML, images, etc)
* `/functions` is gonna contain [Firebase Cloud Functions](https://firebase.google.com/docs/functions/)

(There are also other dot files and configuration files that they added.)

### Domain name

My plan is to create the domain name **`x.hollowverse.com`** and have that be our home for now. Then I want to create a deploy script. 

### Deploy script

The job of the deploy script is the following:

1. Create the JavaScript bundle and put it in `/public`
1. Create `styles.css` and put it in `/public`
1. Copy `index.html` from `/client` to `/public`
1. Run `firebase deploy` (that will upload `/public` and `/functions` to Firebase)

The command to run the deploy script will be **`npm run deploy`**.

I'm thinking of using ShellJS to perform the tasks of the deploy script. I'd like to avoid using Grunt or Gulp if possible.

### Other notes

For now we don't have any Cloud Functions, but I think it should be part of our design guidelines that any code that can be run on the server, should be run on the server. For example, we have code in `vendor.firebase.ts` that handles creation of new user that can be moved server-side. We should move it. I've added that bit to our README.

It should also be noted that the deployment process will change before we launch **v1**. We will need an actual Node server to do server-side rendering and we'll need automatic deployment of `master` through Travis CI...and at that point we'll probably wanna remove `npm run deploy`. And eventually we'll want to use a chat bot to do deployment of branches other than `master`.

### Remaining tasks

- [x] Setup domain name
- [x] Create a build script
- [x] Add deployment instructions to README

### Fixes

Fixes #23, fixes #17, fixes #5, fixes #3

PR #28, opened 2017-04-08
Fix `tsx` linting

wholesomedev:
I didn't setup linting for `tsx` correctly in #22. This PR fixes that.


deltaidea (approved):
Nice style. I'll take it.

PR #32, opened 2017-04-08
Setup dummy Travis CI integration

wholesomedev:
This is just a dummy Travis CI integration. It'll help with Greenkeeper since Greenkeeper requires CI integration with test results to know what to do.

To be clear, I don't think Greenkeeper will go ahead and merge automatically, even if test results are green. It's just the text in the PR is different depending on the results from the CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: node_js
+
+node_js: "node"
+
+install: true
+
+script:
+  - npm run test
+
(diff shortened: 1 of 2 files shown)


wholesomedev:
@deltaidea I'll go cowboy and merge this, since it's just initial config.

deltaidea:
Sneaky you. 🐍

PR #37, opened 2017-04-09
Fix `react-router-dom` component type issue

### LINKED ISSUE #35: TypeScript for React Router is complaining about the `component` attribute
For some reason, on React Router 4, TypeScript doesn't like the following:

```
<Route exact path='/' component={Homepage}/>
```

It's complaining that `Homepage` is not the way it likes. I think this is a bug with the type definitions of React Router 4. As a temporary workaround, I can do

```
<Route exact path='/' component={Homepage as any}/>
```

I think this is a related issue https://github.com/DefinitelyTyped/DefinitelyTyped/issues/13689

wholesomedev:
I think this is how we fix #35

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,8 @@ script:
 
 notifications:
   email: false
+
+branches:
+  only:
+    - master
+    - /^greenkeeper.*$/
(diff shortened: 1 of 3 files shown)


deltaidea (approved):
Fair enough. 👍

PR #39, opened 2017-04-11
Fix Travis deploy

wholesomedev:
@deltaidea Travis CI is not executing the deploy command correctly. It seems that it's bypassing the `npm run` command and executing the content of the script directly, so it is [failing like so](https://travis-ci.org/hollowverse/hollowverse/builds/220738676#L1273).

I decided to remove `npm run deploy` (I don't think we should be running this manually anyway) and change the deployment command in `.travis.yml`.

I also changed `README.md` a bit, and added the Travis CI badge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ script:
 
 deploy:
   provider: script
-  script: npm run deploy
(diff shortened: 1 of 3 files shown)


deltaidea (approved):
Looks good. 👍

PR #46, opened 2017-04-27
Move README.md content to wiki

wholesomedev:
I've moved the content of the README to [the wiki](https://github.com/hollowverse/hollowverse/wiki).


levsthings:
Everything is good except for one point regarding the mentioned resolution size.
> Any screen size with 576px width and higher. We have a mobile-only design without any breakpoints.

As we don't have any breakpoints, I don't see a reason for mentioning a specific resolution. We can just say that we're developing with a mobile-only approach and have no breakpoints for supporting larger devices.

On the other hand, If we want to mention a value and be a bit more specific, the value ` bigger than 576px` would actually exclude most of the smartphones. Instead we can let everyone know how we're developing.

While working on Chrome Developer Tools, there are three smartphone presets available which are called S, M and L. I believe M is the most common one and I find it to be the sweet spot.

In this screenshot, I'm working with M selected.
untitled-1

To get there, one just simply needs to:

1. Select iPhone 6 as a device
2. Set zoom to 100%

We can advise everyone to do same and we'll all see and experience the same thing whilst developing. With the power of `flexbox` and friends, this working resolution should cover all smartphones, phablets and smaller tablets in portrait mode. These account for the most of the mobile market so we'll be safe. 

What do you think @msafih?

wholesomedev:
@levsthings What you said sounds fine to me. Feel free to modify the wiki, and let me know if there are any changes that you'd like me to make to this PR

levsthings (approved):

levsthings:
Everything is fine here, I'll update the Wiki today.

levsthings (approved):

PR #55, opened 2017-04-29
Fix type errors that are failing build

levsthings:
Fixes the type errors so that build goes green before we start working
on packages and dependencies. Also removes an unnecessary Partial.


wholesomedev (approved):
LGTM

PR #62, opened 2017-05-01
Implement circular-dependency-plugin with its custom type definitions

### LINKED ISSUE #52: Detect circular dependencies when bundling
In my experience, circular dependencies cause hard to debug issues. I think we should show a warning when we inadvertently create a circular dependency

This webpack plugin can help https://github.com/aackerman/circular-dependency-plugin

levsthings:
Implements the circular-dependency-plugin with exluding node_modules
and fail on error mode. Also introduces a '/typings/' folder.

Closes #52 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,8 +38,6 @@ build/Release
 node_modules/
 jspm_packages/
 
-# Typescript v1 declaration files
-typings/
 
 # Optional npm cache directory
 .npm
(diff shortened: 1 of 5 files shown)


levsthings:
Type definitions are for:
https://github.com/aackerman/circular-dependency-plugin/blob/master/index.js

wholesomedev (approved):
LGTM!

PR #63, opened 2017-05-02
Break app.tsx to different files and slightly rearrange project structure.

levsthings:
This implements the extraction of Header and removal of Footer from
app.tsx. Also rearranges the folder structure for 'pages', making
necessary changes to TS and SASS imports.


client/src/app/app.tsx
```diff
@@ -1,114 +1,21 @@
...
-    )
-  }
-
-  toggleIsNavMenuOpen() {
-    this.props.setIsNavMenuOpen(!this.props.isNavMenuOpen)
-  }
 }
 
-export const App = connect(mapStateToProps, actionCreators)(AppClass)
+export const App = connect()(AppClass)
```
msafih: If `AppClass` doesn't need any props or actions, why do we still connect it?

client/src/app/header.tsx
```diff
@@ -0,0 +1,97 @@
...
+const actionCreators = pick(actions, [
+  'requestLogin',
+  'requestLogout',
+  'requestUpdateLoginStatus',
+  'setIsNavMenuOpen',
+])
+type ActionCreators = typeof actionCreators
+
+class HeaderClass extends React.Component<ActionCreators & Props, undefined> {
+  componentDidMount() {
```
wholesomedev: Let's put lifecycle hooks in top-level page components. Otherwise things could get messy.

I think it's a good approach to lift lifecycle hooks as high up the component tree as possible.

wholesomedev (changes requested):

client/src/app/header.tsx
levsthings: Good point and that's the guideline we have. This will be addressed in the next commit. 

And about using connect() in `App.tsx`, my intention was to just extract Header as simple as possible and arrange the directory structure, `Header.tsx` and `App.tsx` will go through plenty of changes in the next PR right after this gets merged. Anyway, the next commit which moves the lifecycle method to the highest component addresses that as well.

wholesomedev (approved):

PR #69, opened 2017-05-04
Switch to aphrodite

### LINKED ISSUE #68: Replace SASS with Aphrodite
We've decided that the **CSS in JS** better fits the team (due to fresh React Native experience) and the project. We'll be removing SASS from our stack and replacing it with [Aphrodite](https://github.com/Khan/aphrodite).

While there are more trending (`styled-components`) and known (`react-jss`) libraries out there that enables developers to write CSS in JS, we've decided to go with the simplest tool available for the job.

You can read more about `aphrodite` in these articles:
http://engineering.khanacademy.org/posts/aphrodite-inline-css.htm
https://medium.com/@himrc/why-i-love-css-in-js-aphrodite-vs-radium-b2c9bea9a182

levsthings:
**Part 1 - The Prerequisites** 
- [x] Pull and integrate `aphrodite` to our server-side rendering
- [x] Test and make sure it plays well with existing approach.

**Part 2 - All Good**

- [x] Remove SASS and write new global styles in a single CSS file.

The styles pushed in this PR won't be final for v1, it'll just complete the transition with basic layouts, so that anyone can jump in and add styles using JS later on. Building pages and adding functionality will continue in their own issues.

Closes #68.


levsthings:
**So what's happening here:**

1- Aphrodite is included in our SSR strategy.

2- A file named `app.styles.ts` is added next to `app.tsx` to test some dummy styles. (`.styles` part is my naming convention, we can change how we name style files.) If we continue using this naming convention, we'll have a `<filename>.styles.ts` for each `<filename>.tsx` file just like we had with SASS.

3- All SASS files, packages and webpack build steps are removed.

4- While using aphrodite, we'll have a single plain-old CSS file for our Normalize/Global styles. This file will be a tiny but a very powerful CSS file. (Suggested by the developers of `aphrodite`) This file replaces `globalStyles.sass` and it's named `globalStyles.css`. It exists here: `./src/client/globalStyles.css`

5- This single CSS file is added to the build step, it'll be minified and watched for changes during development. It's channeled to `public/css/styles.css'`.

6- At the moment, this single CSS file contains normalize styles which will most likely stay. At the end of the normalize styles, I've slapped in the current `master` branch's compiled output of SASS. The reason for that is I don't want the project going back to stone age once this gets merged. This huge chunk of CSS will all be deleted as I rebuild the components, getting rid of Bulma selectors that are used in the components and rewriting all styles in JS.

wholesomedev:
This LGTM. We just need to talk about external file vs in-file styles.

wholesomedev (approved):

PR #76, opened 2017-05-11
Build 'Contact Us Form' Page

### LINKED ISSUE #14: Add the ability for visitors to easily send feedback and report issues

JavaScriptErika:
Goals:
- [x] Create a page with its own route
- [x] Include email and message form fields
- [x] Write logic for React to collect data on submit
- [x]  Disable the submit button until the form state is valid
- [x]  Create common `<Input />` component 
- [x] Add/write styles in aphrodite
- [x] Deliver data async (dummy API)
- [ ] Refactor code with (and add) Redux-Form **(in current progress!)**

Closes #14


client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <h1>Contact Us</h1>
+        <form>
+          <label>
+            <input
+              placeholder='Your Email'
+              type='email'
+              id='mail'
+              name='emailAddress'
+              value={p.emailInputValue}
+              onChange={({target: {value}}) => this.handleEmailInputChange(value)}
```
wholesomedev: I think we will be doing this in a lot of our components. Maybe it would make sense for us to create our own common `input` component, call it `Input` and make it delegate to the native `input` component but with an additional prop called `onChangeText` which accepts a callback that expects to be given the text directly?

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return (
+      <div>
+        <h1>Contact Us</h1>
+        <form>
+          <label>
+            <input
+              placeholder='Your Email'
+              type='email'
+              id='mail'
+              name='emailAddress'
```
wholesomedev: We don't need `id` or `name` props

client/src/index.tsx
```diff
@@ -22,6 +24,7 @@ ReactDOM.render(
         {/*/!*<Route path="/create-profile" component={requireUserLogin(CreateProfile)}/>*!/*/}
         {/*/!*<Route path="/search" component={Search}/>*!/*/}
         {/*/!*<Route path="/login" component={Login}/>*!/*/}
+        {<Route path='/contactUs' component={ContactUsForm}/>}
```
wholesomedev: Can you snake-case the path?

client/src/redux/actions.ts
```diff
@@ -35,6 +35,8 @@ export const actions = {
   setIsNavMenuOpen: (payload: boolean) => ({type: 'setIsNavMenuOpen', payload}),
   setLastSearchTerm: (payload: string) => ({type: 'setLastSearchTerm', payload}),
   setCreateProfileUrlInputValue: (payload: string) => ({type: 'setCreateProfileUrlInputValue', payload}),
+  setEmailInputValue: (payload: string | boolean) => ({type: 'setEmailInputValue', payload}),
```
wholesomedev: Why would the email address be a boolean?

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+              value={p.emailInputValue}
+              onChange={({target: {value}}) => this.handleEmailInputChange(value)}
+            />
+          </label>
+          <label>
+            <textarea
+              placeholder='Your Message'
+              id='message'
+              name='userMessage'
```
wholesomedev: We don't need `id` and `name` props here either

wholesomedev (changes requested):

wholesomedev:
@JavaScriptErika I've left some comments on your work so far. Regarding the next step, how about this: instead of doing form validation, how about you disable the submit button until the form state is valid?

A valid form state would be:

1. A valid email address
1. At least one word in the textarea

When these two conditions are met, the submit button is active. Otherwise, it's inactive.

client/src/redux/actions.ts
JavaScriptErika: It should not be. I was experimenting and forgot to remove it before committing, thank you for catching that.

client/src/index.tsx
JavaScriptErika: Yes!

client/src/pages/contactUsForm/contactUsForm.tsx
JavaScriptErika: Great, I'll remove them.

client/src/pages/contactUsForm/contactUsForm.tsx
JavaScriptErika: Removing them now

client/src/pages/contactUsForm/contactUsForm.tsx
JavaScriptErika: I'll get to working on that, it will make the code cleaner, too.

JavaScriptErika:
Thank you for your feedback @msafih . It's interesting you mentioned the disabling/enabling of the submit button because it was a thought that crossed my mind recently.  I also replaced this in my top description of the PR to reflect this instead of form validation.

The validity of the email address is taken care of with html5, and assigning the input as 'type=email', is that fine or are there any additional ways you'd like me to implement that?

client/src/redux/actions.ts
```diff
@@ -37,6 +37,7 @@ export const actions = {
   setCreateProfileUrlInputValue: (payload: string) => ({type: 'setCreateProfileUrlInputValue', payload}),
   setEmailInputValue: (payload: string) => ({type: 'setEmailInputValue', payload}),
   setMessageInputValue: (payload: string) => ({type: 'setMessageInputValue', payload}),
+  setSubmitFormValues: (payload: object) => ({type: 'setSubmitFormValues', payload}),
```
levsthings: Let's not use the object keyword as a type, the object in question needs to be typed. For temporary situations, use `payload: {}` instead.

levsthings (changes requested):

client/src/utils/utils.ts
```diff
@@ -13,4 +13,18 @@ export function pick<T, K extends keyof T>(object: T, predicate: K[]): {[NAME in
   return _pick(object, predicate) as any
 }
 
+export function isValidEmail(email: string): boolean {
```
wholesomedev: FYI: TypeScript can infer the return type of the function, you don't have to specify it. Though specifying it probably enhances readability.

client/src/utils/utils.ts
```diff
@@ -13,4 +13,18 @@ export function pick<T, K extends keyof T>(object: T, predicate: K[]): {[NAME in
...
+  const ePattern = /^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$/
```
wholesomedev: I prefer to avoid abbreviations in code, so `emailPattern` would be a better name here. (Though abbreviations are okay when it's a common conventional one, like `minLength` or `maxLength`.)

The other thing is that this email pattern is too restrictive. If I'm reading it correctly, I don't think it accepts `msafi+hollowverse@msafi.com`, which is the email I use for this project :)

It also doesn't accept non-English characters.

I think a better way to validate email is something like:

```
anyCharacterEvenNonEnglish@anyCharacterEverNonEnglish.AnyCharacterEvenNonEnglish
```

client/src/utils/utils.ts
```diff
@@ -13,4 +13,18 @@ export function pick<T, K extends keyof T>(object: T, predicate: K[]): {[NAME in
...
+  return ePattern.test(email)
+}
+
+export function hasSentence(message: string): boolean {
+  const minLength = 19
+  if (message.trim().length !== 0 && message.length > minLength) {
```
wholesomedev: This logic would accept:

```js
hasSentence('hi                                   '); // true
```

wholesomedev (changes requested):

client/src/redux/actions.ts
wholesomedev: Don't we have linting against that? If not, we should.

JavaScriptErika: I'll make that change. @msafih, I did not receive linting errors typing out `object`.

client/src/utils/utils.ts
JavaScriptErika: That's great to know!

client/src/utils/utils.ts
JavaScriptErika: I'll make changes to that variable name, and keep the naming convention in mind for the future!

Thank you for bringing the regex pattern up to my attention - we most certainly need to include non-English characters and characters such as `+` for emails.

client/src/utils/utils.ts
JavaScriptErika: Noooo! haha, ok - I will revisit this logic so that scenario doesn't return true.

client/src/redux/actions.ts
```diff
@@ -35,6 +35,9 @@ export const actions = {
   setIsNavMenuOpen: (payload: boolean) => ({type: 'setIsNavMenuOpen', payload}),
   setLastSearchTerm: (payload: string) => ({type: 'setLastSearchTerm', payload}),
   setCreateProfileUrlInputValue: (payload: string) => ({type: 'setCreateProfileUrlInputValue', payload}),
+  setEmailInputValue: (payload: string) => ({type: 'setEmailInputValue', payload}),
+  setMessageInputValue: (payload: string) => ({type: 'setMessageInputValue', payload}),
+  setSubmitFormValues: (payload: {}) => ({type: 'setSubmitFormValues', payload}),
```
levsthings: Don't forget, the `payload: {}` is only temporary. If you notice, whenever the type is `object` we have to write an interface for it and use that interface there as the type.

client/src/redux/reducers.ts
```diff
@@ -25,7 +25,10 @@ interface IAppState {
   error: HvError,
   isNavMenuOpen: boolean,
   lastSearchTerm: string,
-  createProfileUrlInputValue: string
+  createProfileUrlInputValue: string,
+  emailInputValue: string,
+  messageInputValue: string,
+  submitFormValues: {email: string, message: string}
```
levsthings: Same thing here, the interface you can write may also define the type of `submitFormValues` here.

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,93 @@
+import * as React from 'react'
+import {connect} from 'react-redux'
+import {RouteComponentProps} from 'react-router-dom'
+import {actions} from '../../redux/actions'
+import {State} from '../../redux/reducers'
+import {pick, isValidEmail, hasSentence} from '../../utils/utils'
+
+export interface ContactFormData {
```
JavaScriptErika: Exporting the `ContactFormData interface` temporarily in this file until @levsthings finds a place for it in the `typings` directory.

client/src/utils/utils.ts
JavaScriptErika: @msafih would you prefer if I remove the return type?

wholesomedev: @JavaScriptErika Your call. I don't have an opinion on this.

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react'
+import {connect} from 'react-redux'
+import {RouteComponentProps} from 'react-router-dom'
+import {IContactFormData} from '../../../../typings/typeDefinitions'
+import {Form} from '../../components/form'
+import {Input} from '../../components/input'
+import {actions} from '../../redux/actions'
+import {State} from '../../redux/reducers'
+import {pick, isValidEmail, hasSentence} from '../../utils/utils'
```
JavaScriptErika: Found the linting error and will fix and include in next commit: `import {hasSentence, isValidEmail, pick} from '../../utils/utils'`

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,223 @@
...
+  //   setTimeout(() => {
+  //     p.setIsSubmitPending(false)
+  //   }, 800)
+  // }
+}
+
+export const ContactUsForm = connect<IProps, ActionCreators, RouteComponentProps<any>>(
+  mapStateToProps,
+  actionCreators,
+)(ContactUsFormClass)
```
wholesomedev: @JavaScriptErika I think you did an excellent job in setting the blur state on the input (this exercise gives you insight into what 3rd party libraries like `redux-form` basically do). However, I don't think we should merge this code in this shape. I think it's time we bring in a 3rd party library that will manage these form properties for us. What do you think?

client/src/pages/contactUsForm/contactUsForm.tsx
```diff
@@ -0,0 +1,223 @@
...
+    p.setEmailInputValue('')
+    p.setNameInputValue('')
+    p.setMessageInputValue('')
+    p.setEmailHasBlur(false)
+    p.setNameHasBlur(false)
+    p.setMessageHasBlur(false)
+  }
+
+  onFormSubmit() {
+    fetch('https://jsonplaceholder.typicode.com/posts/1').then((response) => {
```
wholesomedev: It would be cleaner to move this to Redux Sagas

wholesomedev (changes requested):

client/src/pages/contactUsForm/contactUsForm.tsx
JavaScriptErika: Great - I'll look into making that move

client/src/pages/contactUsForm/contactUsForm.tsx
JavaScriptErika: Thank you @msafih! It is very beneficial to gain the insight with this exercise and see what's going on underneath the hood with the logic. I agree with you - let's bring on the 3rd party library!

forabi:
@JavaScriptErika Any progress on this? Do you need any help integrating Redux Form?

wholesomedev:
@forabi Yesterday @JavaScriptErika told me that she's stepping down. She's done a great job with this so far, so we can finish it off when the time is right. I guess for now we can leave it as-is while we work on other areas?

forabi:
@msafih Yes, we can work on this later.

JavaScriptErika:
Thank you @msafih! I am more than happy to answer any questions you or @forabi may have about the form when you guys come back to it.

forabi:
We made a lot of changes to directory structure and the libraries we are using since this PR was started. Let's close this.

PR #78, opened 2017-05-15
Add events to 'Notable Person Page'

### LINKED ISSUE #58: Build 'Notable Person' Page
At the moment, we haven't decided on how other pages (search, create, edit) will interact with the 'Notable Person' page. 

However, we know for sure that we'll have a 'Notable Person' page and we have an early vision of this page which is clearly described in the [product description](https://github.com/hollowverse/hollowverse/wiki/Product-Description). Also an early [mock-up](https://github.com/hollowverse/hollowverse/blob/master/design-docs/mockups/_outputs/MobilePortrait_v01_InAction.jpg) is available. 

Based on this information, we'll start building the 'Notable Person' page and track the initial progress with this issue. The first thing to do is to flesh out the components that belong to this page with some mock data.

Moving forward, we'll start thinking about how to integrate this page.

- [x] Build initial Notable Person page
- [x] Implement Events
- [x] Connect to Firebase backend

levsthings:
- [x] Design and implement events on Notable Person Page
<s>Design and implement "add events" feature on Notable Person Page</s> _Out of v1 scope._
- [x] Connect to Firebase back-end
<s>Make a final design pass</s> _Will be done for all pages & components in one big PR for total branding right before v1_

Closes #58


levsthings:
> - [x] Design and implement events on Notable Person Page
> <s>Design and implement "add events" feature on Notable Person Page</s> _Out of v1 scope._
> - [ ] Connect to Firebase back-end
> - [ ] Make a final design pass

Currently I'm on the last item, I've commented out the unnecessary functionality like login, hamburger. The design of NP page for the pilot v1 is almost ready, what will be the source of our data for the particular pilot notable person?

wholesomedev:
@levsthings I'm thinking we can manually fill out the Firebase database, and load the data on the NPs page?

levsthings:
Sounds good to me, I've transferred the dummy data to Firebase database, we'll pull from there.

client/src/common.styles.ts
```diff
@@ -1,16 +1,33 @@
...
+    '#f5f5f5',
+    '#ffffff',
+  ],
+  backgrounds: [
+    '#f5f5f5',
+    '#f0f0f0',
+    '#fcfcfc',
+    '#ffffff',
+  ],
+}
```
wholesomedev: As we discussed, I think it's better to have a name of the color that's in the array.

client/src/globalStyles.css
```diff
@@ -213,4 +213,4 @@ body, #app, .mainApp {
     transform: translate3d(0, 0, 0); }
   to {
     opacity: 0;
-    transform: translate3d(0, 20%, 0); } }
+    transform: translate3d(0, 20%, 0); } }
```
wholesomedev: Are we gonna be able to move these to Aphrodite eventually?

client/src/pages/notablePerson/events.styles.ts
```diff
@@ -0,0 +1,53 @@
...
+    color: typography[1],
+
+  },
+  userAvatar: {
+    borderRadius: '20%',
+    height: '100%',
+    marginRight: '0.2rem',
+    verticalAlign: 'top',
+  },
+})
```
wholesomedev, line 53: Should we start moving some things to `common.styles`? @JavaScriptErika will want to use some of these things in #76 to style it?

client/src/pages/notablePerson/events.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react'
+import {css} from 'aphrodite/no-important'
+import {sortByDescending} from '../../utils/utils'
+import {common} from '../../common.styles'
+import {styles} from './events.styles'
+
+interface IProps {
+  data: {
+    notablePersonEvents: {
+      eventId: number,
```
wholesomedev: I think the word `event` is redundant here.

wholesomedev (changes requested):

client/src/globalStyles.css
levsthings: Yeah, this is the last bit of code which relies on CSS, this will be removed when I start working on transitions.

client/src/pages/notablePerson/events.styles.ts
levsthings: Yeah, I'll export as much as I can for common.styles before she starts using CSS.

client/src/pages/notablePerson/events.tsx
levsthings: All the redundant prefixes will be behind us with the latest schema.

client/src/vendor/firebase.ts
```diff
@@ -81,3 +81,22 @@ export function loginOrRegister(facebookAuthResponse: facebookSdk.IAuthResponse)
     return Promise.reject(errors.firebaseLoginError)
   }
 }
+
+// work in progress =>
+export function getData(child: string): Promise<firebase.database.DataSnapshot> {
+  return new Promise<firebase.database.DataSnapshot>((resolve, reject) => {
+    return firebaseDb.ref().child(child).once(
```
wholesomedev: I think `firebaseDb.ref().child(child).once('value')` will return a promise if you don't give it a callback function. Can we use that syntax and use async/await?

I'd like us to avoid using promises in our codebases and use `async/await` instead since they're cleaner.

I know that my code above is doing this, but that code is pretty bad and we should clean it up.

levsthings:
**Summary:**
- NP page design is ready.
- Data types and schemas are ready both in the codebase and on Firebase.
- We're pulling Events data from Firebase and conditionally rendering.
- Additional data request saga logic is in place for pulling userData for future versions. (For displaying avatar, display name under Events, on Header etc..)
- First iteration of `shadowComponents` for hi-fidelity loading is ready.

I think this is enough for this PR to be reviewed and merged.

**Additional tasks that surfaced while working on this PR will be handled in separate PRs:**
- Work on router per our discussions and pull NP data with route parameters.
- If `shadowComponents` will be kept, send a PR to improve their design and usage.
- Make a final design pass on all pages and components right before we reach v1.
- Rewrite all Firebase functions with `async/await` 
- Rewrite Redux file/folder structure.

client/src/common.styles.ts
levsthings: Updated as per our discussions.

client/src/pages/notablePerson/events.styles.ts
levsthings: I think the container level styles, typography styles and palettes will be enough for the commons for now. I haven't exported anything else because the rest are really specific for Notable Person. As I'm writing my custom CSS solutions, reusability of  minor details shouldn't be an issue. Also in terms of page diversity, we don't have many templates, I'd like to keep top-level styles in `common.styles` and template specific implementations on their respective `.styles` files.

client/src/app/app.styles.ts
```diff
@@ -1,8 +1,12 @@
 import {StyleSheet} from 'aphrodite/no-important'
+import {palette} from '../common.styles'
+
+const {shadesOfWhite} = palette
```
wholesomedev, line 4: Nit: maybe should be called `shadesOfGray` because usually white is thought of as binary and without shades...just a suggestion :)

client/src/app/header.styles.ts
```diff
@@ -1,15 +1,21 @@
 import {StyleSheet} from 'aphrodite/no-important'
+import {palette} from '../common.styles'
+
+const {shadesOfBlack} = palette
```
wholesomedev, line 4: But then there's `shadesOfBlack`, so what's the difference between `shadesOfGray` and `shadesOfBlack` 🤔

client/src/components/shadowComponent.styles.ts
```diff
@@ -0,0 +1,42 @@
+import {StyleSheet} from 'aphrodite/no-important'
+import {palette} from '../common.styles'
+
+export const styles = StyleSheet.create({
+  shadowPhoto: {
+    background: 'gainsboro',
+    width: '200px',
```
wholesomedev: Generally speaking, is it possible to avoid fixed widths and heights and use percentages (I'm wondering if such approach is even possible)?

The other thing, why does the shadow photo have potentially different width and height compared to the actual photo?

client/src/components/shadowComponent.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <div className={css(eventStyles.userContainer)}>
+            <span className={css(npStyles.notablePersonLabel, styles.shadowContent, styles.shadowUserContent)} />
+          </div>
+        </div>,
+      )
+    )
+  }
+}
+
+export const ShadowComponent = ShadowComponentClass
```
wholesomedev: I think it would be simpler to create a `notablePersonShadowComponent.ts` that lives alongside `notablePerson.ts`. I don't see a need a higher abstraction yet.

client/src/pages/notablePerson/notablePerson.tsx
```diff
@@ -1,58 +1,87 @@
...
   render() {
     return (
-      <div className={css(common.page)}>
-        <div className={css(styles.notablePersonTitleContainer)}>
-          <img className={css(styles.notablePersonPhoto)} src={dummyData.notablePersonPictureUrl}/>
-          <div className={css(styles.notablePersonText)}>
-            <h1 className={css(styles.notablePersonTitle)}>Religion, politics, and ideas of...</h1>
-            <h2 className={css(common.titleTypography, styles.notablePersonName)}>{dummyData.notablePersonName}</h2>
-            {this.renderLabels()}
+      this.renderNotablePerson()
```
wholesomedev: `render` here just calls `renderNotablePerson`, so maybe rename `renderNotablePerson` to `render` and delete the other `render`?

client/src/redux/sagas.ts
```diff
@@ -83,4 +85,23 @@ function* requestUpdateLoginStatus() {
...
+    yield put(actions.setNotablePerson(firebaseResponse))
+  } catch (error) {
+    throw error
+  }
+}
+
+function* requestUserData(action: IAction<string>) {
+  try {
+    const firebaseResponse = yield firebase.getData(`/users/${action.payload}`)
+    console.log(firebaseResponse)
```
wholesomedev: We're not gonna have any `console.logs` in production, right?

wholesomedev (changes requested):

client/src/app/header.styles.ts
levsthings: Yeah, categorization via the root color is not perfect like this because the categorization I made within the arrays are based on usage not on the root color. That's why I had them named as `background` (white and darker) and `typography` (black and lighter) before. You think we should go back to that? Let's clarify this before I refactor it again.

client/src/components/shadowComponent.styles.ts
levsthings: This CSS used for drawing a precise circle and things such as avatars, profile photos are defined with px values.
The values of shadow components will be different than the originals because they're mostly `spans` doing magic, they don't have nice normalized margins in contrast to real markup with things like `h1, h2, img` etc..

client/src/redux/sagas.ts
levsthings: Oh forgot to clean this function because it was unused.

client/src/components/shadowComponent.styles.ts
levsthings: Also note that we're designing for the smallest screen so we're fully safe in terms of responsiveness.

levsthings:
Requested changes are ready, the only thing remains is the naming of the color arrays.

client/src/app/header.styles.ts
levsthings: untitled-1

client/src/pages/notablePerson/notablePerson.tsx
```diff
@@ -1,58 +1,79 @@
...
+              {this.renderLabels(labels)}
+            </div>
           </div>
+          <Events data={events}/>
         </div>
-      </div>
-    )
+      )
+    } else {
+      return <ShadowComponent type='NotablePerson'/> // Loading? Grayed Out Components? WIP.
```
wholesomedev: The type `NotablePerson` here is PascalCase. I think we should keep PascalCase for TypeScript types and classes and use camelCase for other things.

But anyway, if this is a shadow component specifically for the notable person, you don't need to give it a type, right? You can just render it.

wholesomedev (dismissed):
I had one small non-blocking comment. The rest LGTM!

wholesomedev:
Looks like there are linting issues.

levsthings:
@msafih Yesterday I noticed that there were some new linting rules while working on HollowBot (It uses the most recent `commonpkg`). It started asking for alphabetical imports and couple of other rules which we didn't have on Hollowverse. I thought it was something to do with my settings there. Today, the same issue is happening on Hollowverse. 

Build fails due to:
`interface name must start with a capitalized I`
`Import sources within a group must be alphabetized`

Strange thing is I didn't see these linting errors locally, I had to update dependencies. Did you update these rules yourself or is there something unintended going on?

P.S. I've fixed both of the issues as linter requested, the question is just for clarification. Other than that, the PR is good to go.

wholesomedev (approved):
LGTM

PR #84, opened 2017-05-22
Upgrade commonpkg

wholesomedev:
This PR upgrades commonpkg. 

From now on whenever we want to update common configs, we need to do the following:

1. Add changes to hollowverse/common-config repo
1. Bump the version of hollowverse/common-config repo (not sure if this is really necessary, perhaps we should try without and see if it works)
1. Merge hollowverse/common-config changes to its master branch
1. In this repo, run the command `yarn commonpkg`

...and we'll get the latest changes from common-config.

At some point I'll probably need to document these steps some where. For now, we're all on the same page I think so we're good 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "server/firebase": "yarn client/build && firebase serve",
     "lint": "tslint '**/*.t{s,sx}' -e '**/node_modules/**' --fix --project ./tsconfig.json --type-check",
     "precommit": "lint-staged",
-    "test": "yarn lint && yarn client/build"
+    "test": "yarn lint && yarn client/build",
+    "commonpkg": "yarn add hollowverse/common-config && commonpkg"
   },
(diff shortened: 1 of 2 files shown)


levsthings:
LGTM but this needs to be in wiki as a future reference. We'll definitely remember this but in case someone else is delegated to do this update in the future, they need to be linked these instructions.

levsthings (dismissed):

wholesomedev:
The greenkeeper PR caused a conflict with the `yarn.lock` file. I resolved the conflict by merging master into this PR and regenerating the `yarn.lock` file, but I think what I did brought in a new typing package or something that broke our build. We need to resolve that before I can merge this PR

wholesomedev:
And actually, Travis CI is broken with this PR for some reason. I'll try to close it, delete it, and push again.

PR #94, opened 2017-05-25
Refactor actively used promises in the codebase

levsthings:
I'll be refactoring all the promises that are actively wired and used by the project.


client/src/vendor/facebook.ts
```diff
@@ -1,42 +1,52 @@
...
+  }
 }
 
-export function logout() {
-  return new Promise<facebookSdk.IAuthResponse>((resolve) => {
-    FB.logout((response) => {
-      resolve(response)
+export async function logout(): Promise<facebookSdk.IAuthResponse | void> {
+  try {
+    return await FB.logout((response) => {
```
wholesomedev: Does this syntax actually work? If so, why don't you use the same with the other functions above?

levsthings: All of them work like a charm, the only problematic one is the `login` function for now so it still uses the promise syntax. that one needs more work. (Also realized that I forgot to push my latest save.)

wholesomedev: @levsthings Why is the `login` one different, what's problematic about it?

levsthings: Unlike all the other functions in `facebook.ts`, in `login` we rely on `resolve/reject` to prevent a login loop. Check out the latest version. I still need to convert the callback to a promise and await that.

levsthings: The common practice for times when you need to convert a callback to a promise with `async/await` is as follows:

```
function getQuote() {
  var quote;
 
  return new Promise(function(resolve, reject) {
    request('http://ron-swanson-quotes.herokuapp.com/v2/quotes', function(error, response, body) {
      quote = body;
 
      resolve(quote);
    });
  });
}
 
async function main() {
  var quote = await getQuote();
  console.log(quote);
}
```

My implementation in `5484bd2` is similar but I'm not extracting the promise to a separate function, I'm handling everything in the same block.

levsthings: I have updated the implementation in `09fbc96` to be the same as the example I've pasted in the previous comment.

The reasons for extracting the promise into another function:
1- More common way of doing things.
2- Async/Await block stays synchronous looking.
3- `Firebase.ts` also have similar implementations but with longer code blocks, they'll benefit from extracted promises, I want to keep the style consistent.

wholesomedev:
Hey @levsthings Is this ready? What are the next steps on it?

levsthings:
@msafih I forgot to request a review on this, could you take a look at the last iteration? I don't have any other ideas for this PR, I'll leave the decision up to you for any potential next steps.

wholesomedev:
I would suggest that we create a utility function that promisifies the methods of the Facebook SDK and then convert all those functions into async await.

We might lose some type-safety, but we can figure out how to improve that later.

levsthings:
At the moment there are two Facebook SDK methods that returns promises. If I try to extract promises like I did in the first iteration, I feel like we'll go full circle back. Can you please give me an example of the utility function you wanted using `login` as an example?

wholesomedev:
@levsthings 

I made this quick and dirty, I can clean it up later if you want, but it conveys the idea I think:

```js
  function giveMeCallback(callback) {
    setTimeout(() => {
      callback('this is the results')
    }, 1000)
  }

  // giveMeCallback((results, err) => {
  //   if (err) {
  //     console.log(err)
  //   } else {
  //     console.log(results)
  //   }
  // })

  function promisify(func) {
    return function() {
      return new Promise((resolve, reject) => {
        func((results, err) => {
          if (err) {
            reject(err)
          } else {
            resolve(results)
          }
        })
      })
    }
  }

  const giveMeCallbackPromise = promisify(giveMeCallback)

  giveMeCallbackPromise().then((results) => {
    console.log(results)
  })
```

levsthings:
This was a good example and if you check my last commit, you can see how I implemented it. It works without an issue however the `async/await` I've used there became redundant again. Any ideas on this?

I like the approach of node style "promisify" utilities but we still haven't been able to find a place for `async/await` :grin:

levsthings:
In general I really like node style "promisfy" utilities but it looks weird here.

https://github.com/hollowverse/hollowverse/blob/refactor-used-promises/client/src/vendor/facebook.ts

Can you check this out and let me know if this is similar to what you had in mind? The problem with this is `async/await` is still redundant in this approach. `console.log()` on the results works but sagas don't work as expected in this setup and we lose the types. Am I missing something?

client/src/vendor/facebook.ts
```diff
@@ -1,42 +1,94 @@
...
+    const response = promisifier(FB.getLoginStatus)
+    const status = response().then((results: facebookSdk.IAuthResponse) => {
+      return results
     })
-  })
+    return status
+  } catch (err) {
+    throw err
+  }
 }
```
wholesomedev: @levsthings This doesn't look quite right. The idea is to do something like this:

```js
FB.getLoginStatus = promisify(FB.getLoginStatus)
// FB.login = promisify(FB.login)
// FB.logout = promisify(FB.logout)
// Etc...

async function getLoginStatus() {
  try {
    const loginStatus = await FB.getLoginStatus()
    
    return loginStatus
  } catch(e) {
    throw e
  }
}
```

We are converting FB SDKs methods to promises so that we can use them inside async/await. Promises can be used in async/await syntax. Callbacks cannot.

wholesomedev (changes requested):

client/src/vendor/facebook.ts
levsthings: Oh now I see.

levsthings:
@msafih Could you please review the final status of this for me?

Here's what happened so far: 

All of FB methods other than FB.init are promisifed with our utility function, I've tested all of them, login/logout/getLoginStatus functionality works as intended. FB.init doesn't play well with the promisification, so I'm calling that synchronously.

Let me know what you think but don't merge it even if you like everything because I need to comment out the login/logout button before merging as we don't need it for v1.

client/src/vendor/firebase.ts
```diff
@@ -14,44 +14,51 @@ export const firebaseApp = firebase.initializeApp(firebaseConfig)
...
-      return Promise.reject(errors.firebaseLoginError)
-    })
+        return firebaseAuth.signInWithCredential(credential)
+      })
+      .catch(() => {
+        return Promise.reject(errors.firebaseLoginError)
+      })
+  } catch (err) {
+    throw err
+  }
```
wholesomedev, line 36: Can this function be written to utilize async/await more?

```js
export async function login(payload) {
  try {
    await new Promise((resolve) => {
      const unsubscribe = firebaseAuth.onAuthStateChanged((firebaseUser) => {
        unsubscribe()
        
        resolve(firebaseUser)
      })
    })
    
    const credential = firebase.auth.FacebookAuthProvider.credential(payload.authResponse.accessToken)
    
    await firebaseAuth.signInWithCredential(credential)
  } catch (err) {
    throw err
  }
}
```

...but let's forget about the async/await issue for a second. There seems to be something wrong with this function. The call to `firebaseAuth.onAuthStateChanged`, which gives us the `firebaseUser`, we don't seem to be doing anything with this `firebaseUser`. (I know this is my code, so I screwed up.) 

I think it maybe possible to rewrite this function to even be the following...

```js
export async function login(payload) {
  try {
    const credential = firebase.auth.FacebookAuthProvider.credential(payload.authResponse.accessToken)
    
    await firebaseAuth.signInWithCredential(credential)
  } catch (err) {
    throw err
  }
}
```

...right?

client/src/vendor/firebase.ts
```diff
@@ -14,44 +14,51 @@ export const firebaseApp = firebase.initializeApp(firebaseConfig)
...
-    )
-  })
+        (error: firebase.FirebaseError) => {
+          reject(error)
+        },
+      )
+    })
+  } catch (err) {
+    throw err
+  }
```
wholesomedev, line 61: Can this function be rewritten as such?

```js
  export async function userExists(user) {
    try {
      const usersReference = firebaseDb.ref('users')
      const userSnapshot = await usersReference.child(user.uid).once('value')
      const userExists = userSnapshot.val() !== null
      
      return userExists
    } catch (err) {
      throw err
    }
  }
```

client/src/utils/utils.ts
```diff
@@ -17,4 +17,18 @@ export function sortByDescending<T>(object: {[index: number]: T, length: number}
...
+        if (err) {
+          reject(err)
+        } else {
+          resolve(results)
+        }
+      })
+    })
+  }
+}
+
```
wholesomedev, line 33: Are we going to add types to this function?

client/src/redux/sagas.ts
```diff
@@ -69,8 +69,7 @@ function* requestLogout() {
 
 function* requestUpdateLoginStatus() {
   try {
-    yield facebook.initSdk()
-
+    const sdk = yield facebook.initSdk()
```
wholesomedev, line 72: If `facebook.initSdk()` returns `void` and it's not a promise or async/await, then this `sdk` variable will always be `undefined` and `yield` is not really doing anything. This could cause a race condition where we execute code assuming the SDK has been initialized but it hasn't. Can we implement something like [this solution](https://stackoverflow.com/a/11274607/604296) in the Facebook SDK init function to eliminate any possible race condition?

client/src/vendor/firebase.ts
```diff
@@ -62,23 +69,27 @@ export function createUser(user: IUser) {
...
+            return undefined
+          }
+        })
+    } else {
+      return Promise.reject(errors.firebaseLoginError)
+    }
+  } catch (err) {
+    throw err
   }
 }
```
wholesomedev, line 94: Can this function also be rewritten to utilize async/await?

wholesomedev (changes requested):
I've left a few comments on the Facebook functions and took the liberty to comment on Firebase functions too. Hope that's ok.

PR #104, opened 2017-05-31
Refactor routes

### LINKED ISSUE #101: Refactor routes
As discussed, routes will be refactored so that the we're able to consume route parameters to display the NP via their `id`.

Reference:
https://reacttraining.com/react-router/web/example/ambiguous-matches

levsthings:
Checklist

Refactor routes so that: 
- [x] Either `/:id` or `notable-person/:id` loads the NP page with the corresponding NP data.
- [x] If root level `/:id` is used, it doesn't collide with other root level pages like `/contact` etc.
- [ ] The saga which returns NP data also is able to reject the request based upon for `id`s that don't exist.
- [ ] NP page pushes the user to `/` or `/404` if such `id` doesn't exist via a lifecycle method.

Closes #101


levsthings:
```typescript
ReactDOM.render(
  <Provider store={store}>
    <ConnectedRouter history={history}>
      <App>
        <Route exact path='/' component={Homepage}/>
        <Route exact path='/notable-person' component={NotablePerson}/>
        {/*/!*<Route path='/create-profile' component={CreateProfile}/>*!/*/}
        {/*/!*<Route path="/create-profile" component={requireUserLogin(CreateProfile)}/>*!/*/}
        {/*/!*<Route path="/search" component={Search}/>*!/*/}
        {/*/!*<Route path="/login" component={Login}/>*!/*/}
      </App>
    </ConnectedRouter>
  </Provider>,

  document.getElementById('app'),
)
```
@msafih This is the current Router setup and I need to clarify some things before I move on. The main question is: why are all routes designed to be children of `<App>` component? `<Link>`ing between components doesn't work as I expected in this setup. So I'd like to know if there's a specific reason behind this layout.

wholesomedev:
@levsthings There's no specific reason. I just thought that that's the way to do it with React Router v4.

Regarding this...

> Either /:id or notable-person/:id loads the NP page

Why do we need two URLs that load the same content?

levsthings:
@msafih I didn't mean both of them, meant one or the other. Leaving the `:id` at root level is still experimental for us, we have to make sure that it works fine, if not we can move it to `np/:id`.

levsthings:
@msafih Regarding the layout of the routes, check my latest example. I'd personally lay things out like this, but it's more of a preference thing. 

Also I understood that `Link` isn't working as it used to (in react-router v3) because of other reasons, it has nothing to do with layout.

I need to implement some changes based on the below link for `<Link>` to work. Because right now it only changes the URL it doesn't actually re-render the component based on the changes.

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md#the-solution

PR #109, opened 2017-06-06
"Contact Us" Page Design

levsthings:
This PR tracks the changes made to the design files of the "Contact Us" page.

- [x] Add focus/error/valid state for each input field.
- [ ] Consider adding error messages
- [ ] Design successful send transition and error on submit state.


wholesomedev:
@levsthings We're disallowing form submission if the input fields are invalid. So, I don't think there would be a need for error state for the input fields.

JavaScriptErika:
Great - I'm looking forward to implementing your design @levsthings! In our current contact form, we don't have an input field for the user's `name` - is that something we'd want to implement or keep to `email` and `message` inputs for now? @msafih  

Also for our button, we can change its opacity to show enabled/disabled to communicate to the user if the form is valid for submission.

levsthings:
@msafih Having a disabled button is OK but it doesn't replace sending visuals to the user with helpers (border colors in this case). We can both have a UI that's able to paint focus,error,success states and a smart button that disables the submission. The button is there to avoid false submissions and the UI state is there to help the user. One doesn't replace the other. 

@JavaScriptErika I think for now you can lower the opacity by 30% for the disabled button, I'll think of a better solution for this later. I have made a very minimalistic design on purpose so that we're able to add extra polish and other features without scrapping anything. Once this baseline design is implemented, we'll improve and polish things in code.

Also, I've added "name" field on purpose because probably we'll fire an auto responder thanking the user for contacting and at some point we'll reply the person, knowing the name of the user is pretty important I'd say.

wholesomedev (approved):
I think we can merge this as is for now

Issue #113, opened 2017-07-01
Restructure redux directory


forabi:
## Current directory structure
The `client/src/redux` dir is currently structured this way:

```
client/src/redux
├── actions.ts
├── reducers.ts
├── sagas.ts
├── selectors.ts
└── store.ts
```
While this structure might be simple, it has a few disadvantages:
1. Things are grouped by type of functionality: when we think of our app, we should think in terms of features, not in terms of whether that piece of functionality is a saga or a selector, ...etc.
2. All actions, sagas and reducers must be fetched immediately as the store is used. This becomes less efficient as the store grows. Some libraries (like [`redux-observable`](https://redux-observable.js.org/)) allow adding reducers to the store dynamically, and this structure prevents doing that.
3. It becomes harder to organize and debug as more actions and reducers are added.

## Proposed structure
I'm proposing the following structure:
```
client/src/store
├── features
│   ├── auth
│   │   ├── actions.ts
│   │   └── reducer.ts
│   └── search
│       ├── actions.ts
│       └── reducer.ts
├── helpers.ts
├── index.ts
└── types.ts
```

Note that the `redux` directory is renamed to `store` to avoid conflicts with the `redux` library when #112 is done.

Reducers and selectors are co-located in each feature's `reducer.ts` file. Selectors are just helpers that extract data from the part of the state that reducer handles, so having the reducer and the corresponding selector in the same file makes them easier to maintain and reason about.

`index.ts` is just `store.ts` renamed. It is more convenient and idiomatic:

```typescript
import store from 'client/src/store'; // Resolves to client/src/store/index.ts
```
instead of
```typescript
import store from 'client/src/store/store';
```
The `types.ts` and `helpers.ts` files will be discussed in another issue.

Issue #111, opened 2017-06-29
Improve network and runtime performance


forabi:
## Rationale
The app is currently bundled as a single 800KB JS bundle, that contains CSS styles, routes and all vendor libraries.

So hollowverse is designed as mobile-only web app. We need to optimize for this use case. Mobile devices have different performance characteristics compared to desktop computers. Desktop computers today are powerful and have plenty of computing power. On the other hand, mobile devices (especially Android devices, which comprise at least 80% of mobile devices today) have limited CPU resources. The single-core performance of the top high-end Android devices is way below a typical desktop computer. Single-core performance is probably the most relevant performance metric for JavaScript apps, considering the single-threaded nature of the language.

JavaScript needs to be:
* Fetched
* Parsed
* Compiled

So loading 800KB of JavaScript not only impacts the network time of the app (time to first byte), but it also delays time to first meaningful paint, which might be the important performance metric for mobile web apps. Users expect some content to be displayed within 3 seconds of first visit.

To put that into perspective, consider [the following chart](https://mobile.twitter.com/HenriHelvetica/status/877924754195324928):

> Parse times of 1MB of #javascript of various devices as mentioned by @tkadlec #FluentConf 
> ![dc8d2xbxuaitnnq](https://user-images.githubusercontent.com/1012761/27686625-4bd3a1c2-5cdc-11e7-8da4-37f77998b806.jpg)

So we are pretty close to 1MB of JS. To make matters worse, using CSS-in-JS means that styles are embedded in JavaScript. So instead of fetching the styles and JS in parallel, the browser has to wait until JS is fetched, parsed and compiled, additionally delaying time to first paint.

Here is how our current production bundle breaks down:

![screenshot_20170629_144916](https://user-images.githubusercontent.com/1012761/27687116-3b5bfe96-5cde-11e7-8c87-7431693f8650.png)

[`stats.json` file](https://www.dropbox.com/s/oh8tefig4xrczrl/stats.json?dl=0) can be viewed using https://chrisbateman.github.io/webpack-visualizer/.

* 96% of our code is vendor libraries.
* 24% of our code is React and React DOM.
* We are bundling the entire lodash library, even though we are only including the following functions: `pick`, `pickBy` and `sortBy`.
* We are including a huge number of polyfills by requiring `babel-polyfill`, even if not all of them are required. Browsers have [pretty good support for ES6](https://kangax.github.io/compat-table/es6/).
* We are using CSS-in-JS.
[...]

PR #114, opened 2017-07-02
Get project to build with new Webpack config and start refactoring

forabi:
So as discussed in #111, #113, this PR introduces major changes to the build process and directory structure.

As this PR introduces lots of changes, I made sure to provide descriptive commit messages for each commit and also added extensive comments to `webpack.config.js` and most of the changed files, but here is a short summary of the changes:

1. Significantly improved build process for development and production: all of the changes discussed in #111 are incorporated and in addition Webpack is now configured with Hot Module Replacement (HMR) for development environments so code changes are immediately detected and you do not have to reload the page. Full details are described in the commit message for 8c5608c30a46713b8039eaabcefcb63f71bb18bf.
2. Update developer tooling: use [prettier](https://github.com/prettier/prettier) to automatically format JS, TS and CSS files. It is executed as a git hook before every commit. TSLint now uses [ts-config-pretterier](https://github.com/alexjoverm/tslint-config-prettier) so it does not conflict with prettier and does not complain about things prettier can fix automatically.
3. Restructure redux dir as discussed in #113.
4. Enable TS strict option, and fix a lot of issues in order to get the project to build.
5. Better type definitions for components wrapped in `connect()`.
6. Use `React.PureComponent` every where instead of `React.Component`. Pure components are more efficient and do not break any functionality.
6. Upgrade all dependencies to latest versions.

There are still some issues related to TSLint and the project won't build unless linting is disabled, which can be done with the new `NO_LINT` flag:

```
NO_LINT=1 yarn client/dev
```

Make sure to update dependencies before running the command.


wholesomedev (commented):
Awesome work!

Just a few comments/questions:
* Why do we need to separate `react.js` and `vendor.js`?
* We don't need `stylelint` now, but I guess it'll come in handy if we switch to plain CSS in the future?
* We have a guideline of strictly not using JS anywhere, but I saw your explanation for why using JS makes sense for config files. In that case, can you please update the guidelines?

I'm okay with merging these changes. However, I see that there are some configurations and code that were added as preemptive measures. As a general approach, I would recommend that we start with the simplest code for our immediate needs and as our needs evolve, we can continue to redesign, refactor, and improve.

wholesomedev (dismissed):

wholesomedev:
The CI needs to be fixed :)

forabi:
@msafih The reason for two separate chunks is that the vendor code may change more frequently compared to react libraries (react, react-dom, react-redux). So say if we bring in a new helper function from lodash, `react.js` won't change so the browser won't have to redownload this file again. It basically boils down to how frequently we expect vendor vs react code to change.

forabi:
@msafih We are currently using some global css in `client/src/globalStyles.css` so that's why I added stylelint. Should I remove it?

wholesomedev:
@forabi Ah, I see. No, I think it's fine to keep it. 

With the preemptive code I was mainly referring to some code in `env.js` like `isInferno` and some of those checks are used in `webpack.config.js` and they don't do anything as far as I could tell, like `isES5`.

Speaking of the variable name `isES5`, it doesn't comply with the naming guidelines we have. Could you fix that please?

forabi:
@msafih I fixed two issues blocking CI but it is still failing as I still have to fix TS lint issues. I also fixed the inconsistent camel casing. It takes some getting used to 😅.

wholesomedev:
Yeah, it doesn't always feel natural. It is more consistent though. 😄 

Once this passes the CI, we can merge it.

forabi:
@msafih So all blocking issues are now resolved but the build is failing because of the enforced size limits on webpack assets. How would you like to handle that? Should I disable these limits by prefixing the build command with `NO_PERF_CHECKS=1`?

wholesomedev (approved):
LGTM

Issue #112, opened 2017-07-01
Use absolute imports with configured base dir


forabi:
We are currently using the standard relative imports like this:

```typescript
import { HvError } from '../../../typings/typeDefinitions';
```

* This is quite inconvenient to write as we have to look for the location of the current file and the location of the file to be imported. 
* It also makes it harder to refactor code or move files around as we would need to change the import path string.

TypeScript allows [configuring a base directory](https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url) from which absolute imports are resolved, so if we configure our base dir to be the root directory of out project, the import statement above becomes much simpler:

```typescript
import { HvError } from 'client/src/typings/typeDefinitions';
```

We might also be able to set the base directory to be `client/src`, so it becomes even simpler:

```typescript
import { HvError } from 'typings/typeDefinitions';
```

This way we do not have to think about where this file or the imported file is located, and we can easily move files and reorganize the directory structure without having to fix our imports.

Note that this also needs to be configured in Webpack with [`resolve.modules`](https://webpack.js.org/configuration/resolve#resolve-modules), as Webpack will also have to resolve the imports after being processed by TypeScript.

PR #115, opened 2017-07-02
Allow non-relative imports from local modules with custom TS baseUrl

forabi:
As discussed in #112. I originally suggested we use a `tsconfig.json` for client, server and root where client and server extend from the root. This PR takes a different approach but it basically does the same. The difference is described in the commit message for ba2c09e below.


forabi:
Rebased on master to resolve conflicts

wholesomedev (approved):
LGTM

PR #116, opened 2017-07-03
Refactor promises

forabi:
This should close #94. Also a race condition is fixed when initializing Facebook SDK in authentication saga.

### DIFF
--- a/client/src/constants/errors.ts
+++ b/client/src/constants/errors.ts
@@ -1,13 +1,34 @@
-import { HvError } from 'typings/typeDefinitions';
+import { HvError as AbstractHvError } from 'typings/typeDefinitions';
 
-export const errors: { [code: string]: HvError } = {
-  facebookLoginError: {
-    code: 'facebookLoginError',
-    message: 'We could not complete login to Hollowverse through Facebook',
-  },
-  firebaseLoginError: {
-    code: 'firebaseLoginError',
-    message:
-      'We could get you logged-in successfully. Please try again, and we will be on the case!',
-  },
+export const messagesByCode = {
+  facebookLoginError:
+    'We could not complete login to Hollowverse through Facebook',
+  firebaseLoginError:
+    'We could get you logged-in successfully. Please try again, and we will be on the case!',
 };
+
+type ErrorCode = keyof typeof messagesByCode;
+
+/**
+ * A custom error class for app-specific errors
+ * like authetnication and data fetching errors.
+ *
+ * It is always a good practice to throw actual
+ * Error objects so we can get stack traces.
+ */
+export class HvError extends Error implements AbstractHvError {
+  name = 'HollowverseError';
+  code: ErrorCode;
+  message: string;
+}
+
+/**
+ * A helper function to create an custom error
+ * instance given a valid error code
+ */
+export function makeError(code: ErrorCode): HvError {
+  const error = new HvError(messagesByCode[code]);
+  error.code = code;
+
+  return error;
+}
--- a/client/src/constants/errors.ts
+++ b/client/src/constants/errors.ts
@@ -1,25 +1,8 @@
-import { HvError as IHvError } from 'typings/typeDefinitions';
+import { ErrorCode } from 'typings/typeDefinitions';
 
-export const messagesByCode = {
+export const messagesByCode: Record<ErrorCode, string> = {
   facebookLoginError:
     'We could not complete login to Hollowverse through Facebook',
   firebaseLoginError:
     'We could not get you logged-in successfully. Please try again, and we will be on the case!',
 };
-
-export type ErrorCode = keyof typeof messagesByCode;
-
-/**
- * A custom error class for app-specific errors
- * like authentication and data fetching errors.
- *
- * It is always a good practice to throw actual
- * Error objects so we can get stack traces in cases
- * where we throw an error and forget to catch it somewhere
- * in our code.
- */
-export class HvError extends Error implements IHvError {
-  name = 'HollowverseError';
-  code: ErrorCode;
-  message: string;
-}
--- a/client/src/constants/errors.ts
+++ b/client/src/constants/errors.ts
@@ -7,7 +7,7 @@ export const messagesByCode = {
     'We could not get you logged-in successfully. Please try again, and we will be on the case!',
 };
 
-type ErrorCode = keyof typeof messagesByCode;
+export type ErrorCode = keyof typeof messagesByCode;
 
 /**
  * A custom error class for app-specific errors
@@ -22,15 +22,4 @@ export class HvError extends Error implements IHvError {
   name = 'HollowverseError';
   code: ErrorCode;
   message: string;
-}
-
-/**
- * A helper function to create an custom error
- * instance given a valid error code
- */
-export function makeError(code: ErrorCode): HvError {
-  const error = new HvError(messagesByCode[code]);
-  error.code = code;
-
-  return error;
 }
--- a/client/src/constants/errors.ts
+++ b/client/src/constants/errors.ts
@@ -1,4 +1,4 @@
-import { HvError as AbstractHvError } from 'typings/typeDefinitions';
+import { HvError as IHvError } from 'typings/typeDefinitions';
 
 export const messagesByCode = {
   facebookLoginError:
@@ -18,7 +18,7 @@ type ErrorCode = keyof typeof messagesByCode;
  * where we throw an error and forget to catch it somewhere
  * in our code.
  */
-export class HvError extends Error implements AbstractHvError {
+export class HvError extends Error implements IHvError {
   name = 'HollowverseError';
   code: ErrorCode;
   message: string;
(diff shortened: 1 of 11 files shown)


client/src/constants/errors.ts
```diff
@@ -1,13 +1,34 @@
...
-    message:
-      'We could get you logged-in successfully. Please try again, and we will be on the case!',
-  },
+import { HvError as AbstractHvError } from 'typings/typeDefinitions';
+
+export const messagesByCode = {
+  facebookLoginError:
+    'We could not complete login to Hollowverse through Facebook',
+  firebaseLoginError:
+    'We could get you logged-in successfully. Please try again, and we will be on the case!',
```
wholesomedev: This should say "We could not"? Missing "not".

client/src/constants/errors.ts
```diff
@@ -1,13 +1,34 @@
...
 };
+
+type ErrorCode = keyof typeof messagesByCode;
+
+/**
+ * A custom error class for app-specific errors
+ * like authetnication and data fetching errors.
```
wholesomedev: `authentication` is misspelled.

client/src/constants/errors.ts
```diff
@@ -1,13 +1,34 @@
...
+ *
+ * It is always a good practice to throw actual
+ * Error objects so we can get stack traces.
+ */
+export class HvError extends Error implements AbstractHvError {
```
wholesomedev: Can you help me understand what this is doing? What's the benefit of extending `Error`? What's the benefit of using `AbstractHvError`?

client/src/pages/homepage/homepage.tsx
```diff
@@ -14,7 +14,7 @@ import { styles } from './homepage.styles';
 interface StateProps {
   searchInputValue: string;
   searchTerm: string;
-  searchResults?: IAlgoliaSearchResults;
+  searchResults: IAlgoliaSearchResults | null;
```
wholesomedev, line 17: I used `undefined` instead of `null` because I read that I should do that [here](https://basarat.gitbooks.io/typescript/docs/tips/null.html) and it made sense to me.

client/src/constants/errors.ts
```diff
@@ -1,13 +1,34 @@
...
+/**
+ * A helper function to create an custom error
+ * instance given a valid error code
+ */
+export function makeError(code: ErrorCode): HvError {
+  const error = new HvError(messagesByCode[code]);
+  error.code = code;
+
+  return error;
+}
```
wholesomedev: The idea of the `constants` folder is to keep things that are static and constant, like enums. You've added some functions and dynamic code here, so maybe it should be moved out of `constants` some where?

client/src/vendor/firebase.ts
```diff
@@ -14,48 +14,29 @@ export const firebaseApp = firebase.initializeApp(firebaseConfig);
 export const firebaseAuth = firebaseApp.auth();
 export const firebaseDb = firebaseApp.database();
 
-export function login(
+export async function login(
```
wholesomedev: This function is marked with `async` but it doesn't use `await`. Is that okay/does it make sense? If it doesn't make sense, can we lint against such a pattern?

wholesomedev (changes requested):

client/src/utils/utils.ts
```diff
@@ -16,4 +16,20 @@ export function sortByDescending<T>(
   return sortBy(object, iteratee).reverse();
 }
 
+export function promisify<R>(
+  fn: (cb: (result: R, err?: Error | null) => void) => void,
+) {
+  return () =>
```
wholesomedev, line 24: In the future we might want to modify this utility function to allow us to pass arguments. For now this works though since the Facebook methods we're using it with don't accept any arguments before the callback function.

wholesomedev (changes requested):

client/src/store/sagas.ts
```diff
@@ -66,7 +66,7 @@ function* requestLogout() {
 
 function* requestUpdateLoginStatus() {
   try {
-    yield facebook.initSdk();
+    yield call(facebook.initSdk);
```
wholesomedev: `call` is not really necessary here right? Based on what I understand from the documentation of `redux-saga`, `call` is mostly there to help with testing.

wholesomedev (changes requested):

client/src/constants/errors.ts
forabi: Didn't notice that. I just copy-pasted it :)

client/src/constants/errors.ts
forabi: So the reason I used an abstract class is that I didn't want to introduce any implementation code in a file that only contained types (`typings/typeDefinitions.ts`). `HvError` is an implementation of a that type and it also extends `Error` so we can get stack traces when we throw errors and forget to catch them somewhere.

client/src/constants/errors.ts
forabi: I guess this should be moved to `utils/utils.ts`.

client/src/pages/homepage/homepage.tsx
forabi: It's up to you, but if I remember correctly Redux used to throw an error if a reducer returns `undefined` when reducers are first called to initialize state. But it does not seem to throw if the default state is explicitly set to `undefined` in `createStore`. It seemed strange to accept `undefined` for initial state but throw in reducers.

forabi: Should I revert this change?

client/src/vendor/firebase.ts
forabi: Yes it does not make sense after removing the unused login promise. I will check to see if there is a TSLint rule for that.

client/src/utils/utils.ts
forabi: I tried to support arguments but quickly ran into an issue: you cannot annotate variable number of arguments at the beginning of the function, so if we want to have types for `promisify` with arguments, `cb` must be the first argument to the function we want to promisify, which is not the case for most of JS libraries. (See ["variadic kinds"](https://github.com/Microsoft/TypeScript/issues/5453)).

A workaround that [many libraries do](https://github.com/reactjs/reselect/blob/56c18564226dbc9ead3b03925f9df5cc1b63ede6/src/index.d.ts#L30) is to manually declare a reasonable number of signature overrides for functions that accept variable number of arguments  (start with no arguments, declare override with one argument, another override with two arguments...).

client/src/store/sagas.ts
forabi: Is that the only purpose for `call`? I thought that if we `yield` to a promise, the generator won't wait for that promise to resolve, while if we use `call` it triggers the promise and blocks the generator until it is resolved.

client/src/constants/errors.ts
wholesomedev: Yeah, I think that would be better

client/src/pages/homepage/homepage.tsx
wholesomedev: Your call

client/src/utils/utils.ts
wholesomedev: Ah, I see. Okay, in the future we can use the multiple signature declarations if we needed to...

client/src/store/sagas.ts
wholesomedev: Yeah, as far as I understand, that's its main purpose. [This document](https://redux-saga.js.org/docs/basics/DeclarativeEffects.html) says so

client/src/constants/errors.ts
wholesomedev: I don't understand how it _extends `Error`_. 

In plain JavaScript, I understand that to extend an object we can write code as the following

```js
const newError = {...error, stuff: 'foo'}
```

...but I don't see your code extending the native error any where? Can you help me understand that bit?

client/src/constants/errors.ts
forabi: Where do you think the class definition should go if not in `typeDefinitions`? Should I keep it in `constants/errors.ts`?

wholesomedev: I'm okay with it going to `typeDefinitions`. I just don't think we should have dynamic code in the `constants` folder.

client/src/constants/errors.ts
forabi: The code uses `class ... extends` syntax introduced in ES2015. See the "Sub classing with extends" section on [MDN documentation for classes](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes)

forabi: Basically it is equivalent to setting the prototype of `HvError` to `Error`:

```
function HvError() {
}

HvError.prototype = Error;
```

forabi: Just to avoid any confusion, I am not creating a new error _instance_, I'm subclassing the built-in `Error` and creating a new error _class_. Error instances are created when `makeError` is called.

wholesomedev (dismissed):

wholesomedev (approved):

PR #118, opened 2017-07-05
Continue refactoring Redux store

forabi:
* Add missing action types to new store/types file.
* Finish moving all actions to their respective feature directory
* Update action imports in components to use the new actions
* Completely remove old store code
* Simplify single action reducers significantly by introducing a new `handleAction` helper and a reducer map and using redux's `combineReducers`
* Redeclare window in store/store instead of overriding Window type globally.


wholesomedev (approved):
LGTM!

PR #119, opened 2017-07-06
Upgrade to Webpack 3

forabi:
Upgrade to Webpack 3 and enable Module Concatenation Plugin

* Similar to Rollup, this plugin, introduced in Webpack 3, hoists the scopes for modules in the same chunk and eliminates as many require calls as possible.
* This reduces the bundle size and the runtime overhead of webpack, and may even eliminate the runtime in the best case scenario (see ["The cost of small modules" by Nolan Lawson](https://nolanlawson.com/2016/08/15/the-cost-of-small-modules/)).
* This does not work for CommonJS modules as scope hoisting depends on the ability to statically analyze module imports, which can only be performed on ES modules.
* More details: https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b

This is how our bundle sizes compare with Webpack 2:

Bundle      | Webpack 2  | Webpack 3
------------|-------------|----------------
`vendor.js` | 636.9K       | 620K
`react.js`    | 76K            | 71K
`app.js`      | 45.4K         | 38.3K

And this is how the dependency graphs of each chunk compare (view in full size to compare):

Bundle      | Webpack 2  | Webpack 3
------------|-------------|----------------
`vendor.js` | vendor-2  |  vendor-3
`react.js`    | react-2  | react-3
`app.js`      |  app-2  |  app-3

Our app chunk is the most reduced in terms of size and runtime overhead caused by require calls, which makes sense as our code is written with ES modules while most of the 3rd-party code uses CommonJS.

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -540,7 +540,7 @@ const config = {
       new webpack.optimize.OccurrenceOrderPlugin(true),
 
       // Scope hoisting a la Rollup (Webpack 3+)
-      // new webpack.optimize.ModuleConcatenationPlugin(),
+      new webpack.optimize.ModuleConcatenationPlugin(),
 
       // Minification
       ...ifEs5([
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "shelljs": "^0.7.8",
     "style-loader": "^0.18.2",
     "ts-loader": "^2.2.1",
-    "webpack": "^2.2.1",
+    "webpack": "^3.0.0",
     "webpack-dev-server": "^2.5.0"
   },
   "devDependencies": {
@@ -161,6 +161,6 @@
     "typescript": "^2.4.1",
     "typings-for-css-modules-loader": "^1.5.0",
     "webpack-dev-middleware": "^1.11.0",
-    "webpack-hot-middleware": "^2.18.0"
+    "webpack-hot-middleware": "^2.18.2"
   }
 }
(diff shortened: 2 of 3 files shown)


TheLarkInn:
Instead of bundle sizes, would you consider sharing the JS Runtime/Execution times for those bundles. That is where the REAL win is for scope hoisting.

forabi:
@TheLarkInn The app is still tiny, but I ran Chrome DevTools profiler with a clean profile (no extensions) and set CPU throttling to 2x, here are the results for "Evaluate Script" total time for each bundle:

Bundle      | Webpack 2  | Webpack 3 | Notes
------------|-------------|----------------|---------
`vendor.js` | 75.48 ms    | 78.58 ms        |
`react.js`    | 11.09 ms    | 7.08 ms          | (preact and preact-compat)
`app.js`      | 652.40 ms  | 539.02 ms      |

Here is the [webpack 3 config file](https://github.com/forabi/hollowverse/blob/e90bfc2a9560a330b9e69c5998ec9a7a2d28cc39/client/webpack.config.js) used to build the app. The webpack 2 version uses the exact same config (without module concatenation plugin, of course).

All bundles are minified and `NODE_ENV` is set to `production`.
* OS: Linux f-laptop 4.11.6-3-zen #1 ZEN SMP PREEMPT Thu Jun 22 13:47:45 UTC 2017 x86_64 GNU/Linux
* CPU: Intel Core i5 4210U CPU @ 1.70GHz
* Webpack 2 version: 2.6.1
* Webpack 3 version: 3.0.0

My experience with profiling and flame graphs is limited but I'd be happy to perform more audits.

forabi:
@TheLarkInn Here are the stats files for webpack [3](https://www.dropbox.com/s/wo8yx9dafys3s5y/stats-4.json?dl=0) and [2](https://www.dropbox.com/s/wgonbdpbuo2ylsu/stats-3.json?dl=0).

wholesomedev (approved):

PR #120, opened 2017-07-06
Fix runtime error with undefined reducer state

forabi:
* The new handleAction helper now requires passing the initial state of the state entry because Redux does not allow returning undefined from reducers
* To provide better type checking, the state key must be passed as a type parameter to handleAction until TS fixes https://github.com/Microsoft/TypeScript/issues/10717


wholesomedev (approved):
LGTM

PR #121, opened 2017-07-07
Remove normalize CSS from global stylesheet

forabi:
We are using postcss-normalize which automatically adds a custom build of normalize.css based on the browsers we target (as specified in "browserslist" field in package.json)


wholesomedev (approved):
LGTM

PR #122, opened 2017-07-07
Rewrite server

forabi:
* We decided to drop SSR for v1, the old server is no longer compatible with our client.
* The new server also allows custom configuration of `PORT` and `PUBLIC_PATH` via environment variables.
* Webpack can also output assets to a custom build path `BUILD_PATH` environment variable.

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -57,13 +57,13 @@ if (!env.shouldTypeCheck) {
   log('Skipping type checking!');
 }
 
-const CUSTOM_BUILD_PATH = process.env.BUILD_PATH ? path.resolve(process.cwd(), process.env.BUILD_PATH) : undefined;
+const CUSTOM_BUILD_PATH = process.env.BUILD_PATH
+  ? path.resolve(process.cwd(), process.env.BUILD_PATH)
+  : undefined;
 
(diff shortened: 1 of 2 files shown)


client/webpack.config.js
```diff
@@ -237,6 +237,7 @@ const config = {
       ifReact(ifHot('react-hot-loader/patch')),
       ifPreact(ifDev('preact/devtools')),
       ifHot('webpack-hot-middleware/client'),
+      ifProd('regenerator-runtime/runtime'),
```
wholesomedev, line 245: What does this do?

wholesomedev (approved):

client/webpack.config.js
forabi: Looking at the error trace in runtime, it seems to be required by firebase. The reason this was not required before is because we were importing the entire babel-polyfill package so regenerator-runtime was being added implicitly. After the switch to `babel-preset-env` which rewrites imports for "babel-polyfill" to individual polyfill imports based on targeted browsers. Files inside node_modules are not processed with `babel-loader` as it is expected that all vendor code is ES5, so `babel-preset-env` it cannot see that firebase requires regenerator-runtime. So it must be imported explicitly.

PR #124, opened 2017-07-07
Use Docker in CI

forabi:


client/src/app/header.tsx
```diff
@@ -37,17 +37,17 @@ class HeaderClass extends React.PureComponent<ActionCreators & Props, {}> {
   }
 
   renderLoginVariants(): { icon: string; action(): void } {
-    const { props: p } = this;
+    const { loginStatus } = this.props;
```
wholesomedev, line 40: What does TSLint not like about this?

forabi: A new rule is introduced in TSLint 5.5 ([`no-this-assignment`](https://palantir.github.io/tslint/rules/no-this-assignment/)). There error says: "Do not assign members of `this` to local variables". The rationale seems to be that a variable assigned to `this` could be "a symptom of pre-ES6 practices or not managing scope well". It seems to be intended to prevent code like `var self = this;`. But I agree it does not make sense for destructuring, and it looks like it can allow destructuring by setting `allow-destructuring` to `true`.

scripts/travis-deploy.sh
```diff
@@ -0,0 +1,3 @@
+if [[ ($TRAVIS_BRANCH == "master") && ($TRAVIS_PULL_REQUEST == "false") ]]; then
+  yarn deploy -- --non-interactive --token $FIREBASE_TOKEN;
+fi
```
wholesomedev: I don't speak Bash 😢

wholesomedev: ...and probably many more people know JS/TS than Bash.

I prefer we try to push our tools to their limits than follow the popular "use the right tool for the job" advice.

forabi: I'm not a bash expert either 😄, but it's not doing much. Should I revert this?

wholesomedev: Yeah, I think it would be better to keep scripts in TypeScript (or at least JS) 😊

If you think we should allow JS in `scripts/` please update the guidelines 😄

forabi:
@msafih I think this one is finished. The functionality is equivalent to what we had before, but everything is now run inside docker.

wholesomedev (approved):

PR #125, opened 2017-07-09
Write cloud function to fetch data by NP slug

forabi:
This PR introduces a new cloud function, `fetchDataBySlug`, which takes a slug for NP (the part after `/notable-person/`) and fetches the data for that NP. This function also has a client-side equivalent that requests data from that function by sending a GET request with the `slug` parameter as the URL query:

```
GET https://us-central1-hollowverse-c9cad.cloudfunctions.net/fetchDataBySlug?slug=Tom_Hanks
```

The data, as well as the mapping `slug => ID` is already in place for Tom Hanks. The saga `requestNotablePerson` is rewritten to use the new function and the `notablePerson` component is configured properly to read the slug from the URL and dispatch `requestNotablePerson`. So the following route is actually displayed properly:

```
/Tom_Hanks
```

This PR also reduces the `vendor` bundle size from around 620KB to 542KB by importing only the necessary functionality from Firebase (authentication and database).

## Things that still need to be done before merging
* [x] ~Add a polyfill for `fetch()` or use a lightweight wrapper around `XMLHttpRequest` like [`got`](https://github.com/sindresorhus/got).~
* [x] How should we handle CORS? Cloud functions are hosted on `*.cloudfunctions.net` so they belong to an origin different from our origin. To workaround that, we currently allow all CORS requests, but this also allows any other website to call our cloud function. On the other hand, allowing only `*.hollowverse` origin would make local development inconvenient as the app is served from `localhost` and this origin is not allowed, and Chrome does not allow access to CORS responses from `localhost` even if it is specified explicitly in the `Access-Control-Allow-Origin` header.


client/src/index.tsx
```diff
@@ -17,7 +17,7 @@ ReactDOM.render(
     <ConnectedRouter history={history}>
       <App>
         <Route exact path="/" component={Homepage} />
-        <Route exact path="/notable-person/:id" component={NotablePerson} />
+        <Route exact path="/notable-person/:slug" component={NotablePerson} />
```
wholesomedev: We need to put the slug at the root level, so that `https://hollowverse.com/Tom_Hanks` will work.

client/src/store/types.ts
```diff
@@ -120,8 +124,8 @@ export type Dispatch<T extends ActionType> = (
 /**
  * A Generic dispatch function.
  * Used when we do not care about the action type
- * but it still requiers that the type of action is one of
```
wholesomedev, line 123: VS Code should spell check comments 😄

client/src/store/types.ts
```diff
@@ -134,7 +138,7 @@ export type GenericActionCreator = ActionCreator<ActionType>;
 export type Reducer<S> = (state: S, action: GenericAction) => S;
 
 export type ReducerMap<State extends object = StoreState> = {
-  // tslint:disable-next-line no-suspicious-comment
+  [// tslint:disable-next-line no-suspicious-comment
```
wholesomedev: Is this Prettier's doing?

functions/index.js
```diff
@@ -1,8 +1,37 @@
...
+admin.initializeApp(functions.config().firebase);
+
+/**
+ * A cloud function that fetchs data for a notable person by their slug.
+ * The client should make a request to this function's URL with the slug as
+ * a query.
+ * @example `GET /fetchDataBySlug?slug=Tom_Hanks`
+ */
+exports.fetchDataBySlug = functions.https.onRequest((req, res) => {
+  cors(req, res, () => {
```
wholesomedev: I assume Firebase back-end uses this `cors` as an Express middleware? Will this `cors` call be duplicated in many of our Firebase CF calls?

functions/index.js
```diff
@@ -1,8 +1,37 @@
...
+      .then(id =>
+        admin
+          .database()
+          .ref('/notablePersons')
+          .once('value')
+          .then(snapshot => snapshot.child(id).val()),
+      )
+      .then(data => res.json(data));
+  });
+});
```
wholesomedev: Will we convert this file to TypeScript and to all the good stuff we currently do inside `client/`?

client/src/utils/utils.ts
```diff
@@ -47,4 +48,14 @@ export function makeError(code: ErrorCode): HvError {
...
 }
 
+/**
+ * A type guard that guarantees that an Event has a quote, providing
+ * better control flow analysis and type checking for TypeScript.
+ */
+export function doesEventHaveQuote(
+  ev: EventSchema,
+): ev is EventWithQuoteSchema {
+  return (ev as EventWithQuoteSchema).quote !== undefined;
```
wholesomedev: Could you tell me why you need to do `is` and then assert `as`?

wholesomedev:
> Add a polyfill for `fetch()` or use a lightweight wrapper around `XMLHttpRequest` like [`got`](https://github.com/sindresorhus/got).

I'm okay with either approach. But isn't `fetch` available in all the browsers we target?

> How should we handle CORS? Cloud functions are hosted on `*.cloudfunctions.net` so they belong to an origin different from our origin. To workaround that, we currently allow all CORS requests, but this also allows any other website to call our cloud function. On the other hand, allowing only `*.hollowverse` origin would make local development inconvenient as the app is served from `localhost` and this origin is not allowed, and Chrome does not allow access to CORS responses from `localhost` even if it is specified explicitly in the `Access-Control-Allow-Origin` header.

We can whitelist `*.hollowverse` and use [this Chrome extension](https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi?hl=en) for local development. But, you know, I'm okay with whitelisting `*`. We can go with `*` until we see an actual problem.

client/src/utils/utils.ts
forabi: This is an advanced feature of TypeScript, called a "type guard".

The problem with the `EventSchema` is that it is an intersection type, which may or may not have a quote. Trying to access the `quote`, `sourceUrl` or `sourceName` properties is not safe so TypeScript does not allow it. So we need a way to tell TypeScript that we can provide a guarantee that an event is an `EventWithQuoteSchema`. This is what type guards do.

This basically says to TypeScript: if this function returns `true`, then you can assume that `ev` is guaranteed to be an event with a quote, so you can stop complaining about the type not having a `quote`, `sourceUrl` or `sourceName` properties.

Now because TypeScript performs control-flow analysis, it can see that when this type guard is used in an if statement, then the type of event in the if body is guaranteed to be an `EventWithQuoteSchema`, so it can correctly narrow down to this type.

The type cast is needed for the same reason, we cannot access `quote` without assuming that the type is an `EventWithQuoteSchema`.

functions/index.js
forabi: Yes I plan to, I just wanted to make sure it works as this is my first experience with CFs.

forabi:
@msafih It seems to be [supported](http://caniuse.com/#search=fetch) in all evergreen browsers. Are we targeting Safari 9? I remember testing one of my projects on an iPad with Safari 9 and it wouldn't load because `fetch` was not supported.

OK, let's just keep access allowed for all origins for now.

functions/index.js
forabi: Yes it looks like we will have to use it in every HTTP function. In fact, [the documentation](https://firebase.google.com/docs/functions/http-events) has an example that uses the same `cors` module.

client/src/store/types.ts
forabi: Yes, it is clearly a bug. I can fix it and commit with `--no-verify` to disable pre-commit prettier hook.

client/src/utils/utils.ts
wholesomedev: Cool. That makes sense. So what's the difference between using an intersection type and setting `quote`, `sourceUrl` and `sourceName` as optional keys with `?`?

wholesomedev:
Ah, freaking Apple with their browser support. How about we only target the latest Safari version?

client/src/utils/utils.ts
forabi: Using an intersection type means that the type will either have all the three properties or none of them. Making each one optional does not mean that the three properties go together.

forabi: Using an intersection type means that the type will either have all the three properties or none of them. Making each one optional does not mean that the three properties go together.

forabi:
Okay, let's go with Safari >= 10 :smiley:

client/src/pages/notablePerson/events.tsx
```diff
@@ -2,9 +2,8 @@ import { css } from 'aphrodite/no-important';
 import * as React from 'react';
 import { EventSchema } from 'typings/dataSchema';
 import { common } from 'common.styles';
-import { sortByDescending } from 'utils/utils';
+import { sortByDescending, doesEventHaveQuote } from 'utils/utils';
```
wholesomedev: Would be nice to have a lint rule against this

forabi: I tried [this](https://github.com/vrsource/vrsource-tslint-rules#no-duplicate-imports) but it seems to be incompatible with TSLint >= 5.

forabi:
@msafih I think this is ready

client/src/pages/notablePerson/events.tsx
```diff
@@ -22,32 +22,40 @@ class EventsClass extends React.PureComponent<Props, {}> {
...
-            <p>
-              {event.sourceName}
-            </p>
-          </a>
-          <div className={css(styles.quoteContainer)}>
-            <p className={css(styles.quotedText)}>
-              {event.quote}
+      return sortByDescending(events, 'postedAt').map(event => {
+        let quote;
+        if (doesEventHaveQuote(event)) {
```
wholesomedev: I thought some more about this. And I think we should update our Product Description to specify that each event must contain a quote. So we shouldn't have this case where we have an event without a quote.

I'm worried if we allow events without quotes people will just be posting their opinions on things?

functions/src/.gitignore
```diff
@@ -0,0 +1 @@
+yarn.lock
```
wholesomedev: Hmmm, why do we have a `.gitignore` inside `functions/src/`?

wholesomedev (changes requested):
Left a couple of comments, but looks good otherwise 👍

functions/src/.gitignore
forabi: Firebase CF does not use yarn, so package versions may be different from our local version, so I thought we should not lock our CF dependencies as this may give a false sense of security.

forabi: I can also ignore this file from `<root>/.gitignore`.

wholesomedev: Oh, I didn't know it was possible to have several `.gitignore` files in different folders. TIL.

I think maybe it's better to keep one `.gitignore` file in the root?

client/src/pages/notablePerson/events.tsx
forabi: This will make our code simpler. I will rewrite this later today and update the wiki.

forabi: OK, I wrote this code. It turns out quotes are already documented to be required for events. It was an oversight on my part. I was trying to fit the old data into the new model.

wholesomedev (dismissed):

wholesomedev (approved):

PR #126, opened 2017-07-10
Write initial implementation for redirection proxy

forabi:
This server should be listening on the default port instead of the old WordPress server. The WP server should be switched to listen on a different port so the new one can proxy requests to it if the requested NP is not available in the new version yet.

The port numbers are not final and only intended for demonstration.

## Notes
* ~Because type definitions for http-proxy are not available on npm, the types are installed using `typings`, which fetches type files from DefinitelyTyped.~
* ~The command line tool is also added as a dependency so that it can be run with `yarn typings install`.~


redirectionProxy/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const proxyServer = httpProxy.createProxyServer({ target: OLD_SERVER_ADDRESS });
+
+const redirectionMap = new Map<string, string>([['tom-hanks', 'Tom_Hanks']]);
+
+server.get('/:oldUrl', (req, res) => {
+  const oldSlug = req.params.oldUrl;
+  const newSlug = redirectionMap.get(oldSlug);
+  if (newSlug !== undefined) {
+    const url = new URL(newSlug, 'https://x.hollowverse.com');
```
wholesomedev: Let's make the domain `https://hollowverse.com` and get rid of `https://x.hollowverse.com`

redirectionProxy/index.ts
```diff
@@ -0,0 +1,25 @@
...
+const server = express();
+
+// tslint:disable-next-line no-http-string
+const OLD_SERVER_ADDRESS = 'http://localhost:3000';
+
+const proxyServer = httpProxy.createProxyServer({ target: OLD_SERVER_ADDRESS });
+
+const redirectionMap = new Map<string, string>([['tom-hanks', 'Tom_Hanks']]);
+
+server.get('/:oldUrl', (req, res) => {
```
wholesomedev: Not something to worry about for now, but in the future we might want URLs like `/about` and `/contact-us` to resolve to the new Hollowverse. Not sure how that'll work with the current logic, but we can worry about that later...

typings/modules/http-proxy/index.d.ts
```diff
@@ -0,0 +1,252 @@
...
+      hostRewrite?: string;
+      /** Rewrites the location host/ port on (301 / 302 / 307 / 308) redirects based on requested host/ port.Default: false. */
+      autoRewrite?: boolean;
+      /** Rewrites the location protocol on (301 / 302 / 307 / 308) redirects to 'http' or 'https'.Default: null. */
+      protocolRewrite?: string;
+    }
+  }
+
+  export = Server;
+}
```
wholesomedev: All this typings stuff is a bit noisy. Can we get rid of it some how? Can we maybe contribute to `@typings/` or publish it as a separate npm module or something?

redirectionProxy/index.ts
forabi: We can also add those URLs to `redirectionMap` later.

typings/modules/http-proxy/index.d.ts
forabi: I will check if I can contribute to `@types`. As far as I know, `@types/*` are automated and pulled from DefinitelyTyped automatically, but I'm not sure why this is not the case for `http-proxy`. We can also ignore those files and install them as part of installing dependencies with CLI tool. `typings` supports a `typings.json` file similar to `package.json`.

wholesomedev: Yeah, I think we should find out why `@types/http-proxy` is not published and see if we can help there.

forabi: This is now tracked in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/17954

forabi:
This diagram explains my last commit:
redirection proxy

I've tested the proxy locally with the following paths:

```
/tom-hanks => /Tom_Hanks => New app
/Tom_Hanks => New app
/ashton-kutcher => Old app
/ => Old app
```

While the logic is correct, redirection to the old app is failing with `403 Forbidden` and the following error shown in the browser:

```
ERROR

The request could not be satisfied.

Bad request. 
Generated by cloudfront (CloudFront)
```
I think CloudFront may need to be configured in allow this.

forabi:
@msafih This is ready to go. I have another PR ready for the Google App Engine stuff but needs to be rebased after merging this.

wholesomedev (approved):
LGTM

PR #128, opened 2017-07-13
Clean up repo

forabi:
* Remove old server files
* Simplify resolving paths with env variables in webpack and server


wholesomedev (approved):
LGTM

PR #129, opened 2017-07-15
Configure app for Google App Engine

forabi:
* This allows the app to be deployed to GAE using a Docker runtime.
* Next thing to do would be to automate deployment of different branches from CI to different versions on App Engine.


wholesomedev (approved):
LGTM!

PR #130, opened 2017-07-18
Deploy every branch to a corresponding App Engine version

### LINKED ISSUE #82: Automatically deploy any pushed branch to its own environment
The following is not for v1 or in our immediate plans, but it's something that would be cool to have eventually.

---

I think it would be cool if we automatically deploy branches to live servers. 

That is, whenever a branch is pushed, our CI creates an environment on the fly and automatically deploys that branch to the newly created environment (assuming the branch passes `yarn test`).

Then if you wanna visit the live branch, you add a cookie `branch=feature-foo-or-whatever` to your browser and hit `https://hollowverse.com`. When Release Manager sees this cookie in your request, it directs you to the environment for that branch.

Of course, we also need to make sure we clean up these environments when their branches get deleted.

forabi:
* Using the Google App Engine provider for Travis CI, this commit enables deployment of each branch in the hollowverse/hollowverse repo to a dedicated App Engine version with a matching name.
* These versions do not currently receive any traffic. The traffic has to be split manually in the App Engine Console for now.
* A Google Cloud Platform service account key is required to deploy a version. The key allows access to all the resources in the Hollowverse project on GCP.
* The key is encrypted with AES-256-CBC using the `travis encrypt-file` command and can only be decrypted with the private key stored in Travis secure environment variables, which is not publicly accessible.
* This setup limits Travis CI role in our build process to be a CD provider only, so there are no longer a Docker build process or tests to be performed, and the script command does nothing.

## How to handle Cloud Functions?
This PR leaves some open questions with regard to Firebase cloud functions. Although App Engine and Firebase Hosting both support multiple project versions that can be deployed and running simultaneously, Cloud Functions do not support versions. So all the deployed app versions call one version of our `fetchDataBySlug` function, and while this works now, we may need to introduce some breaking change to our function, so one or more version of our app may break, and it may be the default one that receives most of the traffic.

Closes #82


wholesomedev:
> * Using the Google App Engine provider for Travis CI, this commit enables deployment of each branch in the hollowverse/hollowverse repo to a dedicated App Engine version with a matching name.
> * These version do not currently receive any traffic. The traffic has to be split manually in the App Engine Console for now.

Eventually we'll want to be able to route traffic to the right environment using a query string or a cookie. This is a great step in that direction 👍

> * This setup limits Travis CI role in our build process to be a CD provider only, so there are no longer a Docker build process or tests to be performed, and the script command does nothing.

I feel like with this PR we are creating tighter coupling with Travis CI. Instead of this, I wonder if we can achieve the following flow:

GitHub PR sent => Travis CI pulls down the branch and a Docker image and runs a Docker command => Docker runs the tests and deploys the PR to a GAE instance.

In that setup, everything happens inside Docker and Linux. We are not relying on Travis CI's proprietary hooks such as `deploy`, `provider` and what not.

I'm not entirely against relying on proprietary stuff, but in this case I feel executing things inside a Docker container with a plain Linux environment will make our setup simpler and more flexible.

> This PR leaves some open questions with regard to Firebase cloud functions. Although App Engine and Firebase Hosting both support multiple project versions that can be deployed and running simultaneously, Cloud Functions do not support versions. So all the deployed app versions call one version of our `fetchSlugById` function, and while this works now, we may need to introduce some breaking change to our function, so one or more version of our app may break, and it may be the default one that receives most of the traffic.

What about plain Google Cloud Functions (as opposed to _Firebase Cloud Functions_), do we have the same limitation there?

forabi:
> What about plain Google Cloud Functions (as opposed to Firebase Cloud Functions), do we have the same limitation there?

Unfortunately yes, Cloud Functions cannot be versioned. It looks like we'll need to manage versions manually. So here is one way we could work around this limitation:

We define use a new query string parameter (e.g. `fetchDataBySlug?slug=Tom_Hanks&v=2`) and increment the version number every time we need to introduce breaking changes. Inside the function body, we check for the version number and send a response compatible with this version.

forabi:
@msafih I updated the PR to use a new Docker image (https://github.com/forabi/hollowverse-docker). This image installs Google Cloud SDK and deploys the app to a version that matches the branch name. The image has its own Travis CI/CD pipeline so every change to the Dockerfile in the image repo is pulled and a new image tag is deployed to Docker Hub. There is still room for improvement here but the Docker-based deployment pipeline is functional now.

wholesomedev:
>> What about plain Google Cloud Functions (as opposed to Firebase Cloud Functions), do we have the same limitation there?
>
>Unfortunately yes, Cloud Functions cannot be versioned. It looks like we'll need to manage versions manually. So here is one way we could work around this limitation:
>
>We define use a new query string parameter (e.g. `fetchDataBySlug?slug=Tom_Hanks&v=2`) and increment the version number every time we need to introduce breaking changes. Inside the function body, we check for the version number and send a response compatible with this version.

I think this is a reasonable workaround. But I don't feel too good about this approach. I'm beginning to think that despite the awesome benefits of Cloud Functions, with this limitation, it might make more sense for us to not use Cloud Functions, and use our own Node.js implementation instead. What do you think?

> @msafih I updated the PR to use a new Docker image (https://github.com/forabi/hollowverse-docker). This image installs Google Cloud SDK and deploys the app to a version that matches the branch name. The image has its own Travis CI/CD pipeline so every change to the Dockerfile in the image repo is pulled and a new image tag is deployed to Docker Hub. There is still room for improvement here but the Docker-based deployment pipeline is functional now.

This sounds like the right setup. LGTM 👍 

Can you transfer `forabi/hollowverse-docker` to our org and give me ownership of the Docker Hub thing?

.travis.yml
```diff
@@ -1,20 +1,23 @@
...
+          -e PROJECT=hollowverse-c9cad
+          -e IV=$encrypted_744738cd0ff8_iv
+          -e KEY=$encrypted_744738cd0ff8_key
 
-after_success:
-  - docker run -e TRAVIS_BRANCH -e TRAVIS_PULL_REQUEST -e FIREBASE_TOKEN hollowverse/hollowverse /bin/sh -c "cd /hollowverse && yarn travis/deploy"
+        # Use this image for the container
+        forabi/hollowverse-gae
+      ;
+    fi
```
wholesomedev: Indentation in this block is inconsistent, I think.

I mean lines 10-17 should be the same level of indentation, no?

forabi: Yes, I just wanted to make it clear that 10-14 are flags while 17 is the image name.

forabi:
> I think this is a reasonable workaround. But I don't feel too good about this approach. I'm beginning to think that despite the awesome benefits of Cloud Functions, with this limitation, it might make more sense for us to not use Cloud Functions, and use our own Node.js implementation instead. What do you think?

My thoughts too. I think having our own implementation (let's call it the Hollowverse JSON API 😎) eliminates any possibility of compatibility issues because the API and the client are deployed together on App Engine. That is much more maintainable in the long term because if we decide to drop some version we don't have to worry about whether removing that version's logic in our cloud functions is safe or not.

forabi:
@msafih Do you think this still needs more work? I already moved the ownership of the GitHub and Docker Hub repos. We'll need to merge this and make a new branch in the hollowverse repo to see if the deployment works as the encryption keys are not available for Travis PR builds.

The API stuff should probably go in another PR.

wholesomedev:
What is this `gae-client-secret.json.enc` file for?

wholesomedev (dismissed):
One question left as a comment. Otherwise LGTM

forabi:
> What is this gae-client-secret.json.enc file for?

This is an encrypted JSON file that contains the service key required to deploy to App Engine. As our repository is public, anyone can use the key to deploy to our project on App Engine. So it must be encrypted with a private key. The private key components are stored in Travis settings and passed to our Docker container via `$encrypted_744738cd0ff8_iv` and `$encrypted_744738cd0ff8_key` env variables and are only accessible to the repo owners. Inside the Docker container, [the file is decrypted](https://github.com/hollowverse/hollowverse-docker/blob/0cc64a0793652b3b20ea23671136e2da0d227023/Dockerfile-gae#L25) and used to deploy the app.

forabi:
Oops. I just realized I'm not actually using the decrypted key. I'll need to add the login command.

wholesomedev (approved):
LGTM!

forabi:
And it works! 🎉 
screenshot_20170721_202031

wholesomedev:
Woohoo! 🎉🎉🎉

What are the next steps with this? Does it know when to delete branches, er, versions? 😁

forabi:
> What are the next steps with this?

Like I said in the first commit, the deployed versions do not receive any traffic unless we do it manually. It is possible to split traffic from the command line with Cloud SDK, but how much traffic should a particular branch get? This should be our next step.

> Does it know when to delete branches?

I haven't really thought about that. Travis expects that commits are pushed to some branch, but I don't think it does anything when a branch is removed. Maybe this can be achieved with [GitHub web hooks](https://developer.github.com/webhooks/) and a custom server listening to events from GitHub?

PR #131, opened 2017-07-21
Introduce JSON API

forabi:
The new `/api` path exposes a JSON API to use in our client instead of Firebase cloud functions. The rationale for switching to our own API is that Cloud Functions cannot be versioned in the same App Engine deployments can and this could cause some compatibility issues with our different deployed versions.

* `/api/notablePersons/:slug` fetches notable person data by slug and will replace our cloud function `fetchDataBySlug`.
* This also updates the client to use the new JSON API for fetching notable person data instead of the cloud function `fetchDataBySlug`.

## Is this the right approach?
I think we probably should not invest much in the JSON API.

After I wrote the code, I already saw some issues with implementing a JSON API:
* The main reason we introduced the cloud function `fetchDataBySlug` is because we wanted to avoid sending multiple HTTP requests on the client and move that logic to the server. The client would only need to send one request to our Cloud Function and the hosted function will perform the multiple requests and send the response. If we use HTTPS and HTTP/2, we should not worry about sending multiple requests, as HTTP/2 is way more efficient in this scenario. The same connection is used for multiple requests and the requests are multiplexed. So in contrast to HTTP/1, there is not really much overhead caused by creating a new request with HTTP/2.
* We created the `fetchDataBySlug` function to solve the issue of mapping the old URL scheme to the new one (`tom-hanks` => `Tom_Hanks`). So this endpoint should only be temporary while we are moving to the new app.
* If we decide we need an API later, we should probably go with GraphQL instead of a RESTful API. GraphQL is:
  * declarative
  * type safe
  * more efficient as the client only gets the data it asked for, and we don't need to worry about versioning as everything is strictly typed
  * multiple requests can be combined and sent in batches
  * [the Apollo stack](https://github.com/apollographql) plays very well with TypeScript.


client/src/constants/api.ts
```diff
@@ -0,0 +1,4 @@
+// The API will be on 3000 in development. In production, we use
+// `undefined` which means the same URL base of the app
+// tslint:disable-next-line no-http-string
+export const API_BASE = __DEBUG__ ? 'http://localhost:3000' : undefined;
```
wholesomedev, line 4: We need to update the naming section of the guidelines to allow NAMES_LIKE_THIS for constants. They're all over the codebase at this point 😄

server/src/server.ts
```diff
@@ -1,8 +1,12 @@
 import * as express from 'express';
 import * as path from 'path';
 
+import { api } from './api';
+
 const app = express();
 
+app.use('/api', api);
```
wholesomedev, line 8: Neat.

wholesomedev (dismissed):

wholesomedev:
>I think we probably should not invest much in the JSON API.

So are you saying that with HTTP/2 it doesn't matter whether API orchestration happens on the client or the server?

I thought it's better to do API orchestration server-side because this provides a more consistent and lightweight application. The servers are all near each other and have consistent network connection. Clients on the other hand have varying network connection speeds and distances from servers. If the client does the orchestration of API calls, it would take longer than if the server does it. Also the client app (i.e. bundle) size would be bigger than necessary. What do you think?

Regarding GraphQL, I'd be happy to implement it—I heard a lot of great things about it.

forabi:
> The servers are all near each other and have consistent network connection. Clients on the other hand have varying network connection speeds and distances from servers.

Good point. I haven't really thought about that before. What if we use a CDN like [Cloudflare](https://www.cloudflare.com) or AWS Cloudfront? They take care of distributing our app to edge servers closer to the request origin. Another benefit of Cloudflare is that they provide free, automatic SSL certificates and recently they introduced automatic redirection of all HTTP traffic to HTTPS.

We should probably test both approaches in real world and collect some metrics to help inform our decision.

> Also the client app (i.e. bundle) size would be bigger than necessary.

Probably not much bigger if we are still using Firebase for other things. But we need to measure that as well.

client/src/constants/api.ts
forabi: Updated. I know it was not part of the naming conventions but I like to be able to tell apart constant, immutable values (mostly used for configuration options and env variables) from other variables by simply looking at the variable name.

wholesomedev (dismissed):

wholesomedev:
>What if we use a CDN like [Cloudflare](https://www.cloudflare.com) or AWS Cloudfront? They take care of distributing our app to edge servers closer to the request origin.

I understand how static assets can be served from a CDN, but API calls will have to be routed to the original server. I don't think a CDN can help with API calls.

Anyway, like you said, this discussion maybe a bit premature. We can worry about this later when we start noticing a bottleneck in making and creating API calls.

forabi:
Holding off since we decided not to route any pages to the new app

forabi:
This is no longer relevant. We have implemented a GraphQL API instead.

PR #135, opened 2017-07-28
Implement automated SSL certificate generation and renewal with Let's Encrypt

forabi:
This PR introduces a new App Engine service, `lets-encrypt`, which handles generation and renewal of SSL certificates provided by Let's Encrypt.

The Dockerfile uses `certbot`, the recommended Let's Encrypt client to generate the SSL certificate.

As the certificates generated by Let's Encrypt are [domain-validated](https://en.wikipedia.org/wiki/Domain-validated_certificate), we need to prove that we own the domain we are requesting a certificate for. This can be done in one of two ways:
1. Prove that we control the DNS for that domain (this is tricky to implement with AWS managing our domain names).
2. Prove that we can respond to an HTTP request incoming from the certificate authority at a random endpoint on our domain.

This PR uses the HTTP verification method, `certbot` writes the verification files to disk, and we use a static server to host these files at `hollowverse.com/.well-known/`. Let's Encrypt sends a request to `hollowverse.com/.well-known/acme-challenge/<random_string>/` and if we are able to send the correct response (another random string), we are able to prove ownership and we get the certificate.

To redirect traffic to `/.well-known/` to the `lets-encrypt` service instead of our `default` service, we use the `dispatch.yaml` file, which is a way to define custom routing rules for App Engine services.

## Caveats
It turns out Let's Encrypt does not currently support wildcard certificates (*.hollowverse.com), although support is planned for early 2018. So this won't work for subdomains like `beta.hollowverse.com` unless we specify it explicitly during the generation. Thankfully, since most of the process is automated through our CD and `certbot`, it won't be much of a problem since we can configure our Docker container to be aware of the domain name we will use for each deployed version.

But here is the problem:
Even if issue a certificate for `beta.hollowverse.com` explicitly, we won't be able to use the certificate on that domain, because App Engine can only be configured to route traffic on subdomains with wildcards (`*.hollowverse.com`), and only accepts wildcard certificates for wildcard subdomains.

But if we use GAE routing to route users on `hollowverse.com` to different versions our users can still use HTTPS regardless of which version they are seeing.

If we still need a wildcard certificate to implement HTTPS for all subdomains, we can add another wildcard certificate (issued by a different certificate authority). These certificates won't conflict with each other since App Engine can use the appropriate certificate for each domain or subdomain.

We can continue to develop this service until Let's Encrypt supports wildcard domains, and the switch to Let's Encrypt for subdomains should hopefully be seamless.

## TODO
* [x] Add background service (a Linux systemd service or a cron job) to check for expiration at specified intervals and update the certificate automatically.
* [x] Automate the upload of the public and private keys to App Engine via the GCP SDK every time a new certificate is generated.
* [x] ~Configure the `cert.sh` script to generate certificate for the subdomain corresponding to the deployed version.~


wholesomedev:
>This PR introduces a new App Engine service, `lets-encrypt`

~~I'm not clear on this GAE "service" concept. Is the `lets-encrypt` service part of the same server that serves `hollowverse.com`? I guess it is, so does "service" mean a GAE instance?~~

Never mind. I was able to research it and find [this page](https://cloud.google.com/appengine/docs/flexible/nodejs/an-overview-of-app-engine), which explains services sufficiently.

>But if we use GAE routing to route users on `hollowverse.com` to different versions our users can still use HTTPS regardless of which version they are seeing.

Yeah, both `beta` and `master` should be served from `https://hollowverse.com` without the `beta` subdomain.

>If we still need a wildcard certificate to implement HTTPS for all subdomains, we can add another wildcard certificate (issued by a different certificate authority). These certificates won't conflict with each other since App Engine can use the appropriate certificate for each domain or subdomain.

If we serve both `beta` and `master` from `https://hollowverse.com`, then for now we don't need to worry about generating certificates for other subdomains, except for the other feature we talked about which is deploying branches for internal use.

We were planning to access those branches through links such as `some-branch.v.hollowverse.com`. If we can't have HTTPS on that, can we then access deployed branches through cookies or query params? I remember you said that that's possible as well. If that's indeed possible, then we can use it and completely side-step this subdomain certificate issue for now.

----

Another thing with regard to this PR, do you think it would be better or make sense to move all of this Let's Encrypt configurations to its own repo? Just wondering.

If you're gonna keep this stuff here, can we have consistent folder naming either with camelCase or snake-case. Currently we have a mix. Whatever you decide on, please update the guidelines :)

forabi:
> If we can't have HTTPS on that, can we then access deployed branches through cookies or query params? I remember you said that that's possible as well.

Correct. App Engine can use a cookie (named `GOOGAPPUID`) that has a value between `0` and `999` to route users to different versions. We can assign any value to any version of our app.

> Another thing with regard to this PR, do you think it would be better or make sense to move all of this Let's Encrypt configurations to its own repo? Just wondering.

Unfortunately we cannot do that, because we need to deploy three files simultaneously: `app.yaml`, `letsEncrypt/app.yaml` and `dispatch.yaml` so the services and routing configuration get updated at the same time. I just sumbitted a PR to the docker repo (https://github.com/hollowverse/hollowverse-docker/pull/1) to update the deployment command.

> If you're gonna keep this stuff here, can we have consistent folder naming either with camelCase or snake-case. 

Fixed.

wholesomedev:
>Correct. App Engine can use a cookie (named `GOOGAPPUID`) that has a value between `0` and `999` to route users to different versions. We can assign any value to any version of our app.

Okay, hollowbot can let us know what the `GOOGAPPUIDs` of the branches are when it does the deployment

letsEncrypt/env.sh
```diff
@@ -1,3 +1,4 @@
 export DOMAIN='hollowverse.com';
 export EMAIL='msafi@msafi.com';
 export SERVICE_ACCOUNT='travis@hollowverse-c9cad.iam.gserviceaccount.com';
+export CERT_NAME="lets-encrypt-$DOMAIN";
```
wholesomedev: Nice. I think you did that so we can leverage the code for more domains. 👍

letsEncrypt/Dockerfile
```diff
@@ -21,6 +21,9 @@ RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
 
 RUN apt-get update && apt-get install -y certbot nodejs cron at google-cloud-sdk
 
+# Launch at daemon to run tasks scheduled using at
+RUN atd
```
wholesomedev: Ah, good call. I would imagine this would be a hard problem to debug otherwise?

letsEncrypt/env.sh
forabi: Yes. But it turns out we will still need a persistent ID that is generated by GCP. So this is just a display name. To update the same certificate again, we will need to use the certificate ID.

letsEncrypt/Dockerfile
forabi: We will need to ensure that the service is deployed and available to respond to the request incoming to hollowverse.com from the certificate authority. Since we cannot know in advance when the deployment will finish, we just assume that it will take no longer than 30 minutes. Even if the first run fails, the cron job that runs twice a day will generate or update the certificate again as needed.

forabi: I just realized that we may hit an edge case if we are splitting traffic to multiple versions. The request may get routed to another version and may fail. Since we cannot control the cookies set by the certificate authority request, we cannot know for sure that we can deliver the correct response. Any idea how we may deal with this scenario?

forabi: Maybe we can route all traffic to one version (e.g. master) before updating the certificate, and restore the previous routing settings after success?

forabi:
@msafih I've just finished rewriting the script in JS using `shelljs`. Unfortunately I cannot test the script now as Let's Encrypt rate-limits the certificates issued to a specific domain and I already issued too many certificates for hollowverse.com for testing. I'll have to wait for 6 more days to test the JS version. We should not hit this problem in production (unless we deploy the `lets-encrypt` service too many times per week) because `certbot` will first check if a certificate is stored on disk and won't request a new certificate unless one is not found or is about to expire. Even if we hit this problem, the old certificates will continue to work and the script will generate new ones in a few days. I would like to find a way to avoid this problem entirely, maybe something like an external volume in Google Cloud Platform to store the certificates so they are not removed on every deploy.

letsEncrypt/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "lets-encrypt",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Let's Encrypt service for Hollowverse",
+  "main": "index.js",
+  "license": "MIT",
+  "dependencies": {
+    "shelljs": "^0.7.8"
```
wholesomedev: When does the `npm install` command for this `package.json` get run?

letsEncrypt/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "lets-encrypt",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Let's Encrypt service for Hollowverse",
+  "main": "index.js",
+  "license": "MIT",
```
wholesomedev: Can you change this to `Unlicense`?

letsEncrypt/package.json
forabi: Oops. Missed that.

letsEncrypt/package.json
forabi: Updated

wholesomedev (approved):
LGTM

PR #137, opened 2017-08-01
Make sure all forwarded URLs end with / to avoid redirects

forabi:
This fixes an issue where a request to `hollowverse.com/najwa-karam` (without a trailing slash) gets
redirected to `live.hollowverse.com/najwa-karam/`. The reason for this behavior is that WP sends a
`301` redirect to the canonical URL (which is the one ending with a trailing slash) and CloudFront does not follow the redirects.


forabi:
@msafih If we need to deploy this we'll have to revert https://github.com/hollowverse/hollowverse-docker/commit/3a54b503ec333e8dad2675d9345c91832ba9188f because the `lets-encrypt` service and dispatch files are not merged yet.

wholesomedev:
@forabi Okay, we can merge #135 first. I've approved it.

PR #138, opened 2017-08-01
Fix deploy permissions

forabi:
I had to update service account permissions to allow access to App Engine SSL settings. It looks like I removed some permissions required for deployment. This cannot be tested locally or in my fork as the decryption key components are stored in Travis settings for this repo.


wholesomedev (approved):
LGTM

PR #139, opened 2017-08-02
Use default service account for deployment

forabi:
I couldn't determine the exact cause of the problem. I am confident this is a bug in either `gcloud` command line tool or the Roles features of Google Cloud Platform (which is labeled as ALPHA in the web interface).

I used the same service account on my local machine with logging level set to debug. I got this error:

```
DEBUG: HttpError accessing <https://appengine.googleapis.com/v1/apps/hollowverse-c9cad?alt=json>: response: <{'status': '403', 'content-length': '335', 'x-xss-protection': '1; mode=block', 'x-content-type-options': 'nosniff', 'transfer-encoding': 'chunked', 'vary': 'Origin, X-Origin, Referer', 'server': 'ESF', '-content-encoding': 'gzip', 'cache-control': 'private', 'date': 'Wed, 02 Aug 2017 14:33:50 GMT', 'x-frame-options': 'SAMEORIGIN', 'alt-svc': 'quic=":443"; ma=2592000; v="39,38,37,36,35"', 'content-type': 'application/json; charset=UTF-8'}>, content <{
  "error": {
    "code": 403,
    "message": "Operation not allowed",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ResourceInfo",
        "resourceType": "gae.api",
        "description": "The \"appengine.applications.get\" permission is required."
      }
    ]
  }
}
>
DEBUG: (gcloud.beta.app.deploy) Permissions error fetching application [apps/hollowverse-c9cad]. Please make sure you are using the correct project ID and that you have permission to view applications on the project.
```

The description says that `appengine.applications.get` is required to perform the deployment. Looking at the permissions granted to the role assigned to our Travis account, `appengine.applications.get` is clearly granted:

screenshot_20170802_174402

Even assigning "Project Owner" to the account does not seem to fix the issue.

However, using the default service account, which is automatically created for every new project on GCP, seems to be working.


wholesomedev:
I don't think we should merge this because it will cause other issues. Let's see if we can get the permissions working on the other service.

forabi:
> I don't think we should merge this because it will cause other issues.

I don't feel comfortable using the default account for deployment either.

> Let's see if we can get the permissions working on the other service.

The deploy command fails before attempting to deploy anything. I tested with each individual `*.yaml` file separately, with combinations of two, and with all three files together. This is really getting frustrating. I will spend a couple of hours tomorrow double-checking permissions and roles and looking in the documentation for any thing that might cause a problem like this.

PR #141, opened 2017-08-07
Enable TS type checker for cert.js

forabi:
* This already caught a potential runtime error where the certificate ID pattern may be null if the certificate list command returns no certificates found.


wholesomedev (approved):
LGTM

PR #142, opened 2017-08-12
Enable cron in lets-encrypt service

forabi:
After installing `cron`, it will be enabled on next reboot. Since there is no such thing as "rebooting" in Docker container, we need to enable it manually in order to activate scheduled cron jobs.

### DIFF
--- a/letsEncrypt/Dockerfile
+++ b/letsEncrypt/Dockerfile
@@ -46,12 +46,10 @@ ADD . .
 # Schedule renewal using a cron job
 RUN crontab crontab-update-cert
 
-# 1) Start cron and at daemons to enable schedule tasks 
+# 1) Start cron and at daemons to enable scheduled tasks 
 #
 # 2) Schedule certificate generation 15 minutes from now to make sure
 # the service is deployed before running the command
 #
-# 3) Schedule cron jobs to renew the certificate
-#
-# 4) Serve public files to allow Let's Encrypt to verify domain ownership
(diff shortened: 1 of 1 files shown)


wholesomedev (dismissed):

forabi:
@msafih Oops. It turns out that won't work either. [Docker does not keep the processes started at build time running when starting the container](https://stackoverflow.com/questions/37458287/how-to-run-a-cron-job-inside-a-docker-container). I moved the commands to the `CMD` instruction and made sure they are actually running by testing the image locally and checking the process IDs of `cron` and `at` and the command queues using `atq` and `crontab -l`:

```
❯ docker run -i -t --rm 0bfba87d618f /bin/bash
root@368ee605a57b:/# cron start && atd && (echo 'node /cert.js' | at now + 15 minutes)
warning: commands will be executed using /bin/sh
job 1 at Sat Aug 12 00:47:00 2017
root@368ee605a57b:/# atq
1       Sat Aug 12 00:47:00 2017 a root
root@368ee605a57b:/# pidof cron
12
root@368ee605a57b:/# pidof atd
14
root@368ee605a57b:/# crontab -l
0 0,12 * * * node /cert.js
```

Once this is deployed, we can confirm that the script is run at the specified times by SSHing into the `lets-encrypt` instance and checking Let's Encrypt and `cron` logs in `/var/log`.

wholesomedev (dismissed):

wholesomedev (approved):

PR #148, opened 2017-08-18
Update gcloud secrets with new service accounts

forabi:
After contacting GCP support, they suggested re-creating the service account, although I've previously tried that, it didn't occur to me that we should use a new ID after removing the previous account. This behavior is not documented anywhere as far as I know.

This is from the support ticket thread:

> I've previously removed and recreated the service account with the same ID multiple times with no luck. The only service account that seemed to be able to deploy is the default one created by GCP. But we would like to have granular permissions over the account we use for deployment for security reasons.
> Anyway, I've just tried removing the account again and creating a new one with a *different ID* (travis2@hollowverse-c9cad.iam.gserviceaccount.com instead of travis@hollowverse-c9cad.iam.gserviceaccount.com) and this seems to do the trick. However, as far as I know, it is not documented anywhere that once an ID is assigned to a service account, the same ID cannot be used again for a new one if the previous one is removed.

 It could also be that I created the new accounts with my @hollowverse.com account instead of @gmail.com. Anyway, the issue seems to be resolved now. When running the same commands with the new accounts locally, there are no longer any permission errors.

This replaces the old service account keys with new ones for the newly created service accounts, encrypted with the same AES keys.


forabi:
https://github.com/hollowverse/hollowverse-docker/pull/9 should be merged first.

wholesomedev (approved):

PR #161, opened 2017-08-22
Remove letsEncrypt folder from main repo

### LINKED ISSUE #155: Fix Let's Encrypt rate-limiting

### LINKED ISSUE #160: Separate the deployment of Let's Encrypt service
Currently, when we deploy any new client or server code, the `lets-encrypt` service gets deployed as well. This seems inefficient and it slows down deployments (unless we parallelize the deployment I guess?). Anyway, it seems appropriate to separate these two deployments.

forabi:
The code for the Let's Encrypt service is now in a separate repository and has it own CI/CD process [hollowverse/lets-encrypt](https://github.com/hollowverse/lets-encrypt)

Closes #160 
Closes #155


deploy.js
```diff
@@ -0,0 +1,48 @@
...
+  },
+];
+
+async function main() {
+  const code = await executeCommands([
+    () => writeEnvFile('default', shelljs.env, './env.json'),
+    () => decryptSecrets(secrets, './secrets'),
+    `gcloud auth activate-service-account --key-file secrets/gcloud.travis.json`,
+    // Remove Travis key file so it does not get deployed with the service
+    'rm ./secrets/gcloud.travis.json*',
```
wholesomedev: Should we use `shelljs` for this for better cross platform compatibility?

forabi: Actually this is what `executeCommands` does behind the scenes. If the command is a string, it is executed using `shelljs`. Unless you are referring to `shelljs.rm`?

wholesomedev: Yes, I'm referring to `shelljs.rm`, also for the shell expansion of the `*`. Not sure if different shells do this differently.

forabi: Fixed, also fixed in `hollowverse/lets-encrypt`

wholesomedev (approved):

forabi:
Working on this and #162 at the same time is going to add a lot of complexity. I'm closing this in favor of #162. I've already merged the relevant commits and resolved the conflicts.

PR #181, opened 2017-09-14
Deploy to EB from Travis

forabi:
* [x] Add AWS deployment user secrets to Travis settings
  * [x] Pass these secrets to the Docker container
* [x] Update the build environment image (tracked in https://github.com/hollowverse/build-env/pull/25)
* [x] Read environment variables from Travis
* [x] Remove `buildspec.yml`
* [x] Remove the pipeline for `master` in CodePipeline
* [x] Remove CodeBuild project
* [x] Confirm `eb` commands are working
* [x] Upgrade `@hollowverse/common` (tracked in https://github.com/hollowverse/common/pull/17)


wholesomedev:
So you've decided that we should use Travis to deploy to EB?

forabi:
@msafih There are a couple of small issues with CodeBuild and CodePipeline:

* Turns out `CODEBUILD_RESOLVED_SOURCE_VERSION` is some long string that is generated by CodePipeline. It's something like `arn:aws:s3:::codepipeline-us-east-1-622956136362/Hollowverse/hollowvers/FW6KXol.zip/7959a1a89a699a5ab55169badcb0e3f47bc9cd15`. This is what is showing as the branch name in Sumo Logic. It's not nice.
* There seems to be no way to get the branch name from CodePipeline, we will have to create a pipeline for each branch we want to deploy (`master`, `beta`... etc.) and manually set an environment variable in each environment. Travis on the other hand provides this information and we can use it at build time.
* We can build Hollowbot to work with Travis more easily, and even if we decide to let Hollowbot do the deployment of PRs on its own (i.e. without Travis), we will need the EB CLI available in Hollowbot runtime, this PRs assumes that the EB CLI is installed, so we can reuse it.

wholesomedev (approved):

PR #197, opened 2017-09-27
Add X-Hollowverse-Actual-Environment header

forabi:
This header is included with every request and includes the branch and commit ID of the server environment as stored in the environment file written on Travis CI.

This will serve as the source of truth in case a change to release manager mistakenly routes to an unintended version.


wholesomedev (approved):
👍

PR #199, opened 2017-11-02
Implement new page for Tom Hanks

forabi:
* [x] Connect page with production GraphQL API
  * [x] ~Decide on which GraphQL client library to use (Apollo Client vs. Relay)~ [`graphql-request`](https://github.com/graphcool/graphql-request)
* [ ] Check bundle sizes
  * [ ] Create split points with dynamic `import()` where it makes sense
* [ ] Use new favicon
* [ ] Investigate feasibility of server-side rendering (#203)
* [ ] Consider implementing service worker for ["app shell" model](https://developers.google.com/web/fundamentals/architecture/app-shell)
* [x] Import components from [hollowverse/design](https://github.com/hollowverse/design)
* [x] Make sure CORS is allowed for API (hollowverse/api#6)


forabi:
Blocked on hollowverse/api#7

forabi:
@msafih I think we should merge this but disable redirection to the new app so nothing changes for the visitors, but it becomes easier to review and work on #210.

wholesomedev:
@forabi Okay. Is it ready to review and merge?

forabi:
@msafih It's ready now.

client/src/_variables.scss
```diff
@@ -0,0 +1,20 @@
+/**
```
wholesomedev, line 1: Why does this file name begin with `_`?

client/src/assets/quoteArrowPink.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="12" height="10" viewBox="0 0 12 10">
```
wholesomedev: Do we still need these?

client/src/components/FbComments/index.tsx
```diff
@@ -0,0 +1,160 @@
+/**
```
wholesomedev, line 1: So you've decided to go with creating a bunch of `index.tsx` files? 😄 

Fine by me, I trust your judgement.

client/src/components/FbComments/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    new Promise(resolve => {
+      if (MutationObserver === undefined) {
+        resolve();
+      } else if (this.target) {
+        this.commentsObserver = new MutationObserver(mutations => {
+          mutations.forEach(mutation => {
+            if (
+              mutation.attributeName === OBSERVED_FB_ATTR_NAME &&
+              mutation.target.attributes.getNamedItem(OBSERVED_FB_ATTR_NAME)
+                .value === 'rendered'
```
wholesomedev, line 66: Prettier sometimes is way uglier...

client/src/_variables.scss
forabi: Sass files that start with `_` indicate to the Sass compiler that they should not be included in the output. They are generally used for variables and mixins.

client/src/components/FbComments/index.tsx
forabi: I think we'll have to change this for server-side rendering, because Webpack chunks will need a unique name 😃

client/src/assets/quoteArrowPink.svg
forabi: Removed

client/src/components/FbComments/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+        this.commentsObserver.observe(this.target, {
+          childList: false,
+          attributeOldValue: false,
+          attributeFilter: [OBSERVED_FB_ATTR_NAME],
+        });
+      }
+    });
+
+  tryLoading = () => {
+    this.setState({ isLoading: true, hasError: false, timedOut: false }, () => {
```
wholesomedev, line 85: I'm a little concerned about having per-component state. This will make it hard for us in the future to have a central place where we log all the events of our app, do time-travel, etc.

client/src/components/FbComments/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+      const promises = [
+        importGlobalScript(
+          'https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v2.10',
+        )
+          .then(this.observeCommentsRendered)
+          .then(() => {
+            this.setState({ isLoading: false });
+          }),
+      ];
```
wholesomedev, line 94: Very cool!

client/src/components/FbComments/index.tsx
forabi: I think this should later be refactored into a more generic loadable component. But that's a different issue.

I really dislike the idea of having _everything_ in a Redux store. It can sometimes make things more complicated and coupled to Redux even though they can be managed independently. I generally start with managing state at the component level and only move it up to the global store when it's actually needed. In the case of asynchronous loading, we may have multiple components loading at the same time, so how would be identify each one in a global store?

wholesomedev: @forabi I'm okay with keeping this. 

I do think that eventually we may encounter the limitation of this approach though. The main problem is that you can't have visibility from outside into what the component is going through, unless you monkey patch the component prototype and its `setState`, like [this](https://github.com/ryanflorence/react-lumberjack/blob/master/index.js). And because of this, you can't centralize logging and do state time-travel.

Anyway, we're not doing much logging or time-travel for now, so this is fine. We can revisit this in the future again when/if it becomes an issue.

client/src/components/OptionalIntersectionObserver/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+      return import('react-intersection-observer');
+    }
+
+    return Shim;
+  },
+
+  loading: () => (
+    <MessageWithIcon caption="Loading..." icon={<LoadingSpinner />} />
+  ),
+});
```
wholesomedev, line 51: TIL

client/src/helpers/importGlobalScript.ts
```diff
@@ -0,0 +1,29 @@
...
+      alreadyLoaded.delete(url);
+
+      reject(event.error);
+    };
+
+    script.src = url;
+
+    document.body.appendChild(script);
+  });
+}
```
wholesomedev, line 29: Cool stuff!

client/src/helpers/time.ts
```diff
@@ -0,0 +1,7 @@
+export const delay = async (n = 3000, resolveValue?: any) => {
```
wholesomedev: Why is this file called `time`?

client/src/helpers/url.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
wholesomedev: This file name is not descriptive

client/src/helpers/time.ts
forabi: I wanted to group helpers by category. This was intended to contain other helpers to deal with time. Do you have a better name in mind?

client/src/helpers/url.ts
forabi: Suggestion? `prettifyUrl`?

client/src/webpackEntry.ts
```diff
@@ -1,5 +1,5 @@
 // This will be transformed via babel-preset-env to individual
 // polyfill requires as needed by targeted browsers
-import 'babel-polyfill';
-import './globalStyles.css';
+import '@babel/polyfill';
+import './globalStyles.scss';
 import './index';
```
msafih: Well, I'm glad one of the files I created is still around 😂

client/src/webpackEntry.ts
```diff
@@ -1,5 +1,5 @@
 // This will be transformed via babel-preset-env to individual
```
wholesomedev, line 1: Well, I'm glad one of the files I created is still around 😂

client/src/components/FbComments/index.tsx
forabi: @msafih Yeah, I totally agree. Managing the component state internally is like having a black box, we do not know what's going on from the outside. Whenever we need to do something with that state outside the component, it means that the state just does not belong to the component in the first place.

wholesomedev (approved):
Awesome stuff! 👍

client/src/helpers/time.ts
wholesomedev: I see. Would it be better to put each utility function in its own file? Single util files tend to grow unwieldy pretty quickly and in this case the name is not very descriptive.

Perhaps name is `delay.ts`. But doesn't Lodash already have a util function like this?

client/src/helpers/url.ts
wholesomedev: Yeah, `prettifyUrl.ts` sounds good.

client/src/helpers/time.ts
forabi: @msafih Lodash version accepts a callback. This one returns a promise.

PR #200, opened 2017-11-04
Support Docker Compose-based development workflow

forabi:
## Tasks
* [x] Update Wiki to explain how to set up the client and API containers and use a local API instance for the client


forabi:
I've updated the Wiki to describe the new development workflow. I think this is ready. We can make further changes if required in later PRs.

wholesomedev (approved):

PR #201, opened 2017-11-07
Add Sumo secret definition file

forabi:
This is to consolidate how the secrets structure is defined across all our repos, and remove the corresponding sections from the Wiki as it tends to go out of date pretty quickly!


forabi:
I removed the sections describing the structure of the secrets in the wiki (https://github.com/hollowverse/hollowverse/wiki/Managing-Sensitive-Data)

wholesomedev (approved):

PR #210, opened 2017-11-12
Implement Server Side Rendering

### LINKED ISSUE #203: Investigate feasability of implementing server-side rendering
## Why we should implement server-side rendering

* While it's true that SSR is not a panacea and the complexity of implementing it may not be worth it for interactive web apps, Hollowverse is a content-first web app with a few interactions, so it should not depend on JS to show its content. Ideally users should be able to browse the web app even with JavaScript disabled.
* Having the markup delivered and parsed before JS is even fetched will enhance our first meaningful paint considerably
* Having the markup delivered as HTML allows the browser to start parsing and rendering components before the page is finished loading. That's not the case for JS-rendered components.
* We can also push server-rendered markup as it's being produced using streams, this should further enhance our time to first paint.
* SSR should also be beneficial from an SEO perspective.

## Is it possible?

There is no "ultimate" solution to implementing SSR, and it is quite complex to write code that will work in browsers and Node.js, especially when using Webpack-specific syntax for things like file loaders.

I think this is the best time to start looking into this issue because the codebase is still small and there are not much browser-specific code.

## Where we are now

For reference, this is [WebPageTest results](https://www.webpagetest.org/result/171110_25_abcf154044c5010ebcb5986662d65599/) for https://hollowverse.com/tom-hanks?branch=new-app at the time or writing this issue:

screenshot_20171110_144808

(Note that some of the problems in this report do not have anything to do with SSR)

screenshot_20171110_153942

## Relevant Links
* https://github.com/faceyspacey/babel-plugin-dual-import
* https://github.com/faceyspacey/extract-css-chunks-webpack-plugin

### LINKED ISSUE #204: Investigate how to improve caching and compression for static files
[WebPageTest results](https://www.webpagetest.org/result/171110_25_abcf154044c5010ebcb5986662d65599/1/performance_optimization/) show that we are not doing a great job at caching and compressing our Webpack bundles. There is a lot of room for improvement. Although the JS and CSS files are minified, they are not gzipped. They do not have any cache control headers either.

Some possible solutions we could look into:
* Use a CDN, probably Cloudfront, in front of the whole web app, but only cache static assets. This should be done carefully to avoid any possible security problems caused by caching pages with cookies. We should also take into account how this could affect Release Manager routing.
* Use Amazon S3 to store the build artifacts, and Cloudfront in front of the S3 bucket, and point the client to fetch the static files from S3.
* Do all the work on our app server:
  * Customize caching headers on our app server to set `max-age` or `expires` to a reasonably long time. We could also try [`Cache-Control: immutable`](https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/). This tells the browser to store the resource permanently and never ask for it again (not even check if it has expired). This should be safe for our static files as they all contain unique content-based hashes in their names.
  * Use a Node.js module to enable gzip and brotli compression for static files..

forabi:
## How it works
The basic idea for server-side rendering is to generate two Webpack bundles:

* _Client bundle for browsers_
* _Server bundle for Node.js:_ this bundle exports a function that returns an Express middleware, so its type signature is:

    ```typescript
    declare function createServerRenderMiddleware(options: Options): RequestHandler
    ```

    The middleware returned from this function is a standard Express middleware. When it receives a request, let's say for `/Tom_Hanks`, it has to respond with the server-rendered markup of Tom Hanks' page. To be able to do that, it needs to pass the request URL to React Router. The `App` component, which specifies all React Router routes, is wrapped in `StaticRouter`, which is an in-memory router designed for use in Node.js and unit tests.

     ```typescript
     <StaticRouter context={context} location={req.url}>
      <App />
     </StaticRouter>
     ```

    Just like in the browser, the router will look for matching components and render them. From there, it is simply a matter of calling React's `renderToString` to generate the markup.

Things get way more complicated when using Webpack's code-splitting. Until very recently, mixing SSR with code-splitting was considered an almost impossible task.

## Why is it so complicated?
Here are the problems that need to be solved for SSR with code-splitting:

* **Synchronous resolution of components on the server:** to be able to render the page on the server, we need to _synchronously_ resolve all the components that would otherwise be loaded asynchronously on the client.
* **Flushing components required for the initial render to the client**: when JS is loaded on the client, the same components that were used to render on the server must be available immediately so that the initial client render matches the server markup. Otherwise, the client will re-fetch the components asynchronously and throw away all the SSR markup. That's wasteful and might lock up the main thread.
* **Loading the rest of the app on demand:** all other components that were not used on the server, should be loaded asynchronously and only when needed (standard code-splitting).
* **Fetching API data on the server:** in addition to all these problems, there's one important problem to solve when doing SSR: we need to fetch data from the API on the server so that the server-rendering markup contains meaningful content instead of loading placeholders. Without that, SSR would not really be much useful.
* **Chunking CSS and loading it in parallel with JS**: as a bonus, it would be great if we could, instead of bundling the entire app's CSS in one big file, chunk CSS per component and, instead of waiting for JS to be fetched and parsed to reach the CSS require call, load the CSS it _in parallel_ with the JS for that component. Well, it turns out it's possible!

Fortunately, the community has finally found the [code to crack these problems](https://medium.com/faceyspacey/code-cracked-for-code-splitting-ssr-in-reactlandia-react-loadable-webpack-flush-chunks-and-1a6b0112a8b8) individually. Here is a table of the libraries used to solve each problem:

Problem | Library
 ----------|-----------
Synchronous resolution of components on the server | [`react-universal-component`](https://github.com/faceyspacey/react-universal-component)
Flushing components required for the initial render to the client | [`webpack-flush-chunks`](https://github.com/faceyspacey/webpack-flush-chunks/)
Loading the rest of the app on demand | [built-in webpack feature](https://webpack.js.org/guides/code-splitting/)
Fetching API data on the server | [`react-resolver`](https://github.com/ericclemmons/react-resolver)
Chunking CSS and loading it in parallel with JS | [`extract-css-chunks-webpack-plugin`](https://github.com/faceyspacey/extract-css-chunks-webpack-plugin) + `react-universal-component`

[This article](https://medium.com/discovery-engineering/component-level-isomorphic-webpack-code-splitting-b98922382cc1) sums up the complexity of server-side rendering with code-splitting and also goes into detail on the same libraries I used in this PR.

### Directory structure
This PR required changes to the directory structure to be able to more easily connect the server rendering middleware with the client bundle.

Old directory structure:

```
.
├── client
│   ├── dist
│   └── src
└── server
    ├── dist
    └── src
```

New directory strucutre:

```
.
├── dist
│   ├── appServer.js (serves new app, including the SSR middleware and the client bundle)
│   ├── mainServer.js (redirects to new app or old website according to redirection map)
│   ├── client (webpack's client bundle)
│   └── server (webpack's server bundle)
└── src
    ├── app (new app code)
    ├── appServer.ts
    ├── mainServer.ts
    ├── middleware
    │   ├── redirectToHttps.ts
    │   └── security.ts
    └── webpack (webpack configuration files)
        ├── webpack.config.client.js
        ├── webpack.config.common.js
        └── webpack.config.server.js
```

### Developer Experience
This PR does not negatively impact the developer experience. Hot Module Reloading still works as expected and no server restarts are required.

### Other updates in this PR
* Enable gzip and brotli compression for static assets
* Enable caching for static assets, with `immutable` directive
* Add favicons for desktop and different mobile platforms
* Refactor CSS to use CSS modules

## Tasks
* [x] Split webpack configuration into server, client and common
  * [x] Create webpack server entry file
  * [x] Configure hot module reloading for server bundle too
* [x] Extract middlewares to separate folder
  * [x] Extract security middlewares
* [x] Reorganize server code into:
  * [x] **App server**: serves the new, server-side rendered app as well as the client assets
    * [x] Make sure to use the security middlewares for the app server too
    * [x] Serve static assets at `/static/` to more easily control caching headers and compression
    * [x] Enable gzip and brotli compression for static assets
    * [x] Enable caching for static assets for a reasonably long time
      * [x] Use the `immutable` directive for `Cache-Control` header to tell browsers to never check again if the static resource has expired
  * [x] **Main server**: routes requests to the new app or the old website according to the redirection rules
  * [x] Do not start a separate server for the new app, instead export a middleware and use it in the main server
* [x] Use `PureComponent`
* [x] Do not use `index.tsx` for component filenames
* [x] Refactor CSS for CSS Modules
* [x] Make sure `postcss-normalize` is only bundled once
* [x] Replace usage of `react-loadable` with `react-universal-component` where it makes sense
* [x] Write script to build server and client on CI
* [x] Make sure all `nodeExternals` are production dependencies
* [x] Support development workflow with hot module reloading for server-side rendering (server restart and rebuilds should not be required)
* [x] Remove unneeded dependencies
* [x] Comment `webpack.config.server.js` and `webpack.config.client.js`
* [x] Comment `babel.js`
* [x] Follow naming convention for components in `pages`
* [x] Fix imports of SVG files
* [x] Generate and use favicons from new logo
* [x] Handle `react-resolver` errors per-component
* [x] Write and contribute type definitions for:
  * [x] `react-resolver` (https://github.com/ericclemmons/react-resolver/pull/133)
  * [x] `react-universal-component` (https://github.com/faceyspacey/react-universal-component/pull/75)
  * [x] `webpack-flush-chunks` (https://github.com/faceyspacey/webpack-flush-chunks/pull/50)
  * [x] `shrink-ray` (https://github.com/aickin/shrink-ray/pull/17)
  * [x] `webpack-hot-server-middleware` (https://github.com/60frames/webpack-hot-server-middleware/pull/49)
  * [ ] ~`webpack-merge`~ (rewrite in progress https://github.com/survivejs/webpack-merge/pull/95)

Closes #203 
Closes #204 (although we could still investigate the CDN solution in the future)


wholesomedev:
Thanks for writing this explanation. That, along with the excellent article that you’ve linked to, provides a lot information that makes me appreciate how complex SSR with code splitting is. 

It is exciting to be able to deliver an SSR app while being able to fine tune what you ship over the wire. However, it is a little concerning to me how involved this setup is. I hope that it won’t be too difficult to maintain and that it won’t be brittle.

The article you linked talks about React checksum validation. Does Preact have the same issue?

forabi:
> However, it is a little concerning to me how involved this setup is. 

To be honest, I had the same concerns when working on this. However, the packages used to achieve SSR and code splitting are completely decoupled. In fact, `react-univesal-component` can work as a code-splitting helper and does not require a server-side component.

To verify that, I removed one line of code that sets up the server rendering middleware:

```typescript
appServer.use(createServerRenderMiddleware({ clientStats }));
```
and replaced it with a line that serves `index.html` from the client build:

```typescript
appServer.use((_, res) => {
  res.sendFile(path.join(clientDistDirectory, 'index.html'));
});
```

To my surprise, everything worked exactly as expected without SSR: JS bundles were included and loaded, the client performed the request to the API, and the page was rendered exactly like the SSR page. 

It's really impressing that I didn't have to change a single line of code in the web app.

I agree that the setup is pretty involved, but I think what we are building here is the infrastructure and we shouldn't need to make substantial changes in the future.

That said, we will still have to manually test and verify code changes, something we should be doing anyway.

forabi:
@msafih The PR is ready for review! 🎉 

It's also deployed at https://hollowverse.com/Tom_Hanks?branch=new-app

forabi:
With SSR enabled, time to first meaningful paint went from 7.0s to 3.0s on 3G:

screenshot_20171114_233915

screenshot_20171114_233947

And with improved caching configuration, first meaningful paint on repeat views is at 2.0s on a Moto G (gen 4).

package.json
```diff
@@ -51,60 +49,57 @@
...
-    "react-addons-css-transition-group": "^15.6.0",
     "react-dom": "^15.6.1",
     "react-intersection-observer": "^2.0.1",
-    "react-loadable": "^5.3.1",
-    "react-redux": "^5.0.5",
+    "react-resolver": "^3.1.1",
     "react-router": "^4.0.0",
     "react-router-dom": "^4.0.0",
-    "react-router-redux": "^5.0.0-alpha.4",
-    "redux": "^3.7.1",
```
wholesomedev, line 88: Are we removing Redux permanently or is it because it's not being used for now?

I'm seeing things listed under `@types` that are no longer in `dependencies`. Should those be cleaned up?

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -0,0 +1,136 @@
...
+          caption="Are you connected to the internet?"
+          description="Please check your connection and try again"
+          icon={warningIcon}
+          onActionClick={reload}
+        />
+      );
+    }
+
+    const { data } = queryResult;
+    if (!data.notablePerson) {
```
wholesomedev, line 68: Hmmm, are we going to send a 404 from the server in this case?

src/app/helpers/results.ts
```diff
@@ -0,0 +1,7 @@
+export type ErrorResult = { error: Error | any };
+export type SuccessResult<T> = { data: T };
+export type Result<T> = SuccessResult<T> | ErrorResult;
+
+export function isErrorResult<T>(result: Result<T>): result is ErrorResult {
+  return (result as ErrorResult).error !== undefined;
+}
```
wholesomedev, line 7: Can you tell me how you plan to use this in the future other than an ErrorResult type guard? What's the point of the generic `T`?

Can the file name be improved? When I read it I felt it's too ambiguous.

src/app/hocs/AsyncComponent/AsyncComponent.tsx
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  componentDidMount() {
+    this.tryLoading();
+  }
+
+  render() {
+    return this.props.children({ ...this.state, retry: this.tryLoading });
+  }
+}
```
wholesomedev, line 96: I like how you lifted this functionality into an HOC.

src/app/index.server.html
```diff
@@ -10,10 +10,17 @@
...
+      ${app}
     </div>
+    <script>
+      window.__REACT_RESOLVER_PAYLOAD__ = ${data};
+    </script>
+    ${js}
+    ${cssHash}
   </body>
-</html>
+</html>
```
wholesomedev, line 26: What's the difference between `app/index.server.html` and `app/index.html`

src/app/server.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    const icons = iconStats ? iconStats.html.join(' ') : '';
+
+    res.send(
+      interpolateTemplate({
+        // In order to protect from XSS attacks, make sure to use `serialize-javascript`
+        // to serialize all data. `JSON.stringify` won't protect from XSS.
+        // If `data` contains "</script><script>alert('Haha! Pwned!')</script>",
+        // `JSON.stringify` won't help.
+        data: serializeJavaScript(data, {
```
wholesomedev, line 71: How could `data` come to contain `</script><script>alert('Haha! Pwned!')</script>`? 😁

src/app/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../node_modules/@hollowverse/common/tsconfig.json",
```
wholesomedev, line 2: Can you just do `@hollowverse/common/tsconfig.json`?

src/appServer.ts
```diff
@@ -0,0 +1,106 @@
...
+const logger = loglevel.getLogger('Web App Server');
+
+const { APP_SERVER_PORT = 3001 } = process.env;
+
+logger.setLevel(logger.levels.INFO);
+
+const appServer = express();
+
+// Ignore requests for favicons
+appServer.use(noFavicon());
```
wholesomedev: What are we doing with our favicons now? 🤔

I'm seeing a lot of code around favicons. Some here and some in webpack config?

src/mainServer.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (redirectionPath !== undefined) {
+    // /tom-hanks => redirect to Tom_Hanks
+    res.redirect(`/${redirectionPath}`);
+  } else if (newPaths.has(reqPath)) {
+    // /Tom_Hanks => new hollowverse
+    appServer(req, res, next);
+  } else {
+    // /michael-jackson, ashton-kutcher, / => old hollowverse
+    next();
+  }
```
wholesomedev, line 76: So, the URL `Tom_Hanks` is navigated to...

...currently `redirectionMap` is an empty map. When you call `get(reqPath)` on it, you'll always get `undefined`. So the first `if` branch is skipped. `newPaths.has(reqPath)` will return `false`, so the `else if` branch will be skipped. So you go to the `next()`. So how is it that `https://hollowverse.com/Tom_Hanks?branch=new-app` is working?

Never mind. I just looked at the `new-app` branch where the redirection map is not empty 😅

src/webpack/babel.server.js
```diff
@@ -0,0 +1,3 @@
+const { createBabelConfig } = require('./babel');
+
+module.exports = createBabelConfig(true);
```
wholesomedev: What's this file?

wholesomedev (approved):
Epic work here! A lot of the webpack config files were a bit over my head, but this looks awesome!

src/webpack/babel.server.js
forabi: This creates babel configuration for the server bundle.

src/appServer.ts
forabi: Oops! Never mind this. This was a piece of code I copied from https://github.com/faceyspacey/universal-demo. I just removed it.

We are generating favicons for various platforms using https://github.com/jantimon/favicons-webpack-plugin from our logo. This plugin will inject the icon links in the client HTML file (`index.html`), and will also generate a JSON file containing the paths to the generated icons. We use this file to also inject the HTML in the server HTML markup.

src/app/server.tsx
forabi: Well, for Tom Hanks' page there shouldn't be any risk using `JSON.stringify`, but it's still a good practice security-wise. All data should be escaped.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: I thought about this, but I'm not actually sure how to handle response status codes with server-side rendering. I guess we can somehow store this in the `context` object we are passing to React Router.

src/app/index.server.html
forabi: `app.server.html` contains placeholders for the SSR data and markup that should be filled on the server.

`app/index.html` is the regular, empty HTML file we would use for a client-side only app. I kept it in case we decide to implement service workers. The service worker should respond with this file when any page is requested and React Router will look at the actual URL and render the matching page.

package.json
forabi: > Are we removing Redux permanently or is it because it's not being used for now?

It's just because it's not being used now.

I just removed the unused types.

src/app/tsconfig.json
forabi: I just tried to build with `@hollowverse/common/tsconfig.json` but it failed.

src/webpack/babel.server.js
wholesomedev: But I'm not seeing it being used any where?

src/app/helpers/results.ts
forabi: > What's the point of the generic `T`?

This is currently used in `NotablePerson.tsx`. `isErrorResult` needs to accept the type parameter `T` because `Result` is generic with `T` required, and `Result<T>` is generic because we need to express the shape of data returned by the API query.

src/webpack/babel.server.js
forabi: You are right 😅

src/app/helpers/results.ts
wholesomedev: I don't see you passing a value/type for `T` when you call `isErrorResult` in `NotablePerson.tsx`?

forabi: > Can the file name be improved? When I read it I felt it's too ambiguous.

Suggestions?

forabi: > I don't see you passing a value for T when you call `isErrorResult` in NotablePerson.tsx?

TypeScript infers the type of `T` from the type of `queryResult`. So `Result<NotablePersonQuery>` infers `isErrorResult<NotablePersonQuery>` .

wholesomedev: TIL. Can you share a link where I read more about that? When I used TypeScript I always passed the "value" for generics manually?

wholesomedev: > Suggestions?

Don't know. I'm okay keeping this if you want.

forabi: This docs page contains basic information about type inference (although there isn't anything specific to type parameters) https://www.typescriptlang.org/docs/handbook/type-inference.html

The page about generics briefly mentions inference https://www.typescriptlang.org/docs/handbook/generics.html (look for the word infer)

I've become used to this that I don't actually remember how I learned about it. When in doubt about whether the type is inferred I just hover over the code in VS Code.

src/app/hocs/AsyncComponent/AsyncComponent.tsx
forabi: Technically it's not a HOC, although the folder is called `hocs`. It's a component with a render prop, and it's the new hype in React's world! https://cdb.reacttraining.com/use-a-render-prop-50de598f11ce

forabi:
> The article you linked talks about React checksum validation. Does Preact have the same issue?

I assume so. Preact uses a similar diffing mechanism to update the rendered DOM, although the implementation details may be different.

src/app/hocs/AsyncComponent/AsyncComponent.tsx
wholesomedev: Cool stuff! I do agree about all the points stated about the shortcomings of mixins and HOCs. I like the Render Props pattern because it's more transparent!

PR #212, opened 2017-11-22
Add support for event types and labels

forabi:


forabi:
I took the liberty to swap the position of the event labels and the other metadata

screenshot_20171125_124045

I felt this makes it easier to scan the events because the date and source are almost always available and take approximately the same horizontal space across all the events. Also, the labels are not actionable so they do not have to be at a consistent position.

forabi:
hollowverse/api#20 must be merged and deployed in order for CI to pass for this PR.

src/app/components/App/App.global.scss
```diff
@@ -39,7 +38,16 @@ body {
 input,
 button,
 textarea {
-  font-size: 15px;
   font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial,
     sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
 }
+
+a {
+  text-decoration-skip: ink;
```
wholesomedev, line 46: Nice!

src/app/components/Card/Card.module.scss
```diff
@@ -0,0 +1,43 @@
+@import 'variables.scss';
+
+.card {
+  margin: 10px -5px;
+  border: 1px solid rgba(0, 0, 0, 0.08);
+  box-shadow: 0 2px 5px rgba(0, 0, 0, 0.04);
+  border-radius: 3px;
+  padding: 15px;
+
+  &-content > * {
```
wholesomedev, line 10: Who transforms this into `cardContent`, CSS modules or Sass?

src/app/components/Event/Event.tsx
```diff
@@ -1,46 +1,60 @@
...
   } | null;
-  isQuoteByNotablePerson: boolean;
   sourceName: string;
   sourceUrl: string;
+  entityName?: string | null;
+  entityUrl?: string | null;
+  labels: Array<{
+    id: string;
+    text: string;
+  }>;
```
wholesomedev: Hmmm, I got tslint errors when I used this format. It wanted `{id: string; text: string }[]`.

src/app/components/Image/Image.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+          if (errorComponent !== undefined && hasError) {
+            return errorComponent;
+          } else if (loadingComponent !== undefined && isLoading) {
+            return loadingComponent;
+          }
+
+          const { style } = rest;
+      
+          return <div
```
wholesomedev: Is it going to be a problem with styling that this component sometimes returns `img` element, which is `display: inline`, and sometimes returns `div` element, which is `display: block`?

src/app/components/Image/Image.tsx
```diff
@@ -0,0 +1,73 @@
...
+                [classes.isLoading]: isLoading,
+                [classes.hasError]: hasError
+              },
+            ])}
+          />;
+        }}
+      </AsyncComponent>
+    );
+  }
+}
```
wholesomedev, line 75: Love the concept here!

src/app/components/PersonDetails/PersonDetails.tsx
```diff
@@ -16,14 +17,20 @@ export const PersonDetails = ({
...
+    <ul aria-label="Labels" className={classes.personDetailsLabels}>
+      {labels.map(({ id, text }) => (
+        <li className={classes.labelListItem} key={id}>
+          <Label text={text} />
+        </li>
+      ))}
+    </ul>
     {summary && (
       <div className={classes.personDetailsAbout}>
         {summary.split('\n').map((p, i) => <p key={i}>{p}</p>)}
```
wholesomedev: I know this change is not part of this PR, but I just noticed it: can you rename `p` to `paragraph` to avoid confusion? It looks here like you might be using JSX to initialize `p`.

src/app/components/Quote/Quote.module.scss
```diff
@@ -0,0 +1,45 @@
...
+    display: table;
+  }
+
+  &-avatar {
+    flex-shrink: 0;
+    align-self: baseline;
+    position: relative;
+    top: 2px;
+    // stylelint-disable-next-line plugin/no-unsupported-browser-features
+    object-fit: cover;
```
wholesomedev, line 38: TIL

wholesomedev (approved):
Awesome stuff! 👍

src/app/components/Event/Event.tsx
forabi: Yeah, there were two conflicting rules enabled but that was fixed in `hollowverse/common`

src/app/components/Card/Card.module.scss
forabi: CSS modules (webpack's `css-loader`)

wholesomedev:
@forabi Can you send a PR to the design repo with the adjustments you made to the design and CC @Maxgrb on that? I'd like us to keep the `design` repo as the "source of truth" for the design, if possible.

src/app/components/Image/Image.tsx
forabi: You're right. I will change this to `span`

src/app/components/PersonDetails/PersonDetails.tsx
forabi: Fixed

src/app/components/Image/Image.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import * as classes from './Image.module.scss';
+
+/**
+ * This component wraps a regular `<img>` tag and optionally shows
+ * a fallback component while the image is loading or if it has failed to load
+ */
+export class Image extends React.PureComponent<Props> {
+  load = async () => new Promise(async (resolve, reject) => {
+    const img = document.createElement('img');
```
wholesomedev, line 18: @forabi By the way, how does this work with SSR?

forabi: @wholesomedev `<AsyncComponent />` default state is `{ isLoading: false, hasError: false }` and it only starts loading when `componentDidMount` is called. So on the server, where `componentDidMount` is never called, the state will stay `{ isLoading: false, hasError: false }` which is considered equivalent to `hasLoaded` (since `hasLoaded = !isLoading && !hasError`), so the regular `` tag is rendered.

forabi:
> @forabi Can you send a PR to the design repo with the adjustments you made to the design

Sure

PR #215, opened 2017-11-25
Use HTTP Link header to preload static assets

forabi:
This should help browsers start fetching the resources required to render the page sooner.

Reference: https://blog.cloudflare.com/http-2-server-push-with-multiple-assets-per-link-header/

Note that this is not HTTP/2 server push (although the blog post linked above mentions that Cloudflare magically converts those hints to HTTP/2 pushes). This is equivalent to using `<link rel="preload" />` in HTML, but since the hints are in the HTTP headers, they are processed earlier.

In HTTP/2 server push, the server starts pushing the resources before the browser even asks for them. This might be wasteful if the browser already has those resources cached. Although the browser can ask the server to stop pushing the resources, the message may be delayed so the server could have finished pushing all the resources before it got the message to stop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "@types/autoprefixer": "^6.7.2",
     "@types/body-parser": "^1.16.4",
     "@types/domready": "^0.0.29",
-    "@types/express": "^4.0.36",
+    "@types/express": "^4.0.39",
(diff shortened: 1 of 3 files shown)


wholesomedev (approved):
Ah, super cool!

PR #216, opened 2017-11-26
Add Google Analytics script

forabi:


wholesomedev:
Does it make sense to dynamically generate `index.html` and `index.server.html` in order to consolidate their commonalities?

With regard to Google Analytics, 

1) I'm thinking we should consider the new Hollowverse to be a new website and not re-use the old Hollowverse GA account?
2) Is all that's required is just embed the Google Analytics script in our `index.html` file? The new Hollowverse is more like an app/SPA. Doesn't that mean we have to follow the instructions in these articles?

    https://developers.google.com/analytics/devguides/collection/analyticsjs/screens
    https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

forabi:
> Does it make sense to dynamically generate index.html and index.server.html in order to consolidate their commonalities?

I tried. It was somewhat complicated so I thought it's OK to repeat some code here. I may give it another shot.

> 1. I'm thinking we should consider the new Hollowverse to be a new website and not re-use the old Hollowverse GA account?

I'm not sure how to create a new account. I'll check that.

> Is all that's required is just embed the Google Analytics script in our index.html file? The new Hollowverse is more like an app/SPA. Doesn't that mean we have to follow the instructions in these articles?

I didn't know about these instructions before. Looks like they are meant to be followed when the user navigates away from a page to another one where the navigation is handled by JavaScript. Since we only have a single page now, I'm not sure how that applies.

wholesomedev:
>I'm not sure how to create a new account. I'll check that.

I've created a new account. The tracking ID is: `UA-110141722-1`. Let's update this PR before we merge it.

>I didn't know about these instructions before. Looks like they are meant to be followed when the user navigates away from a page to another one where the navigation is handled by JavaScript. Since we only have a single page now, I'm not sure how that applies.

Okay. So I guess we'll have to think about this when our app offers the ability for users to navigate between pages.

wholesomedev (approved):

PR #224, opened 2017-12-12
Little fixes

forabi:
* Use `await` instead of promise chain in FbComments
* Add cite attribute to `blockquote` for better HTML semantics


forabi:
#223 should be merged to make this pass CI

wholesomedev (approved):
👍

PR #225, opened 2017-12-14
Update design, add editorial summary, search functionality and homepage

### LINKED ISSUE #221: Look into how to provide any required attribution for the Wikipedia material that we use

forabi:
* [x] Decide which animation library to use (if any)
* [x] Add `photoUrl` to Algolia index
* [x] Add skeleton component for search results while loading
* [ ] ~Add placeholder for search results page when box is empty~
* [x] Add placeholder for search results page when no results are found
* [x] Add skeleton component for Notable Person page
* [x] Add related people section
  * [x] Add related people to database
  * [x] Add related people to `NotablePerson` type
* [x] Add image attribution link
  * [x] Add image table to database
  * [x] Expose image information via API query
* [x] Make sources collapsable, collapse by default
* [x] Include Algolia logo in search page
* [x] Add home page
  * [x] Search box
  * [x] Add footer
* [x] Add polyfill for `IntersectionObserver`

Closes #221


tslint.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": [
     "./node_modules/@hollowverse/common/tslint.json"
-  ]
+  ],
+  "rules": {
+    "no-backbone-get-set-outside-model": false
```
wholesomedev, line 6: lol

wholesomedev:
@forabi 

Here are a few comments on the UI side so far. I've classified the comments as "Future enhancements" (things that we can address later) and "Priority" (things we should prioritize).

# Future enhancements

* When I click on the back button on desktop, a gray border appears around the back button. That shouldn't happen
* Sometimes image of NP shows the body but not the face of the NP (example: https://hollowverse.com/Jack_Black)
* Nav bar scrolls until out of view then becomes sticky and switches to search bar
* Search text is too small (mismatch with design)
* Input field gets highlighted when I click on the magnifying glass
* Cover image doesn't look good on wider screens. It looks as a repeating pattern. I removed the blurring to check and I saw Steve Austin's repeating dome as a cover image: image
* Quotation symbol is too strong, it clutters up the page
* "Other interesting profiles" section requires side scrolling on narrower displays. Is that what we want?
* Cards don't match design. These cards are elevated and have rounded corners. I think the flat cards are cleaner (and are closer to Facebook's design)
* Summary text is too crammed (needs bigger line-height or letter-spacing perhaps?)

# Priority

* Ryan Dunn page is empty
* Michael Jackson's image is missing
* Some times the Facebook Comments component is stuck on "Loading" (Steps to reproduce: load homepage, search for "st", select Steve Austin, scroll to bottom quickly and tap on one of the other interesting profiles, then scroll to bottom quickly and tap on another interesting profile. I did that a few times and Facebook comments wasn't loading any more) 
* Sources is always open
* On https://hollowverse.com/Chris_Martin page, at the bottom there's a link with the text "his wife". There are missing spaces around that text.

forabi:
> When I click on the back button on desktop, a gray border appears around the back button. That shouldn't happen

screenshot_20171229_141459

Are you referring to this? This is the default "focus ring" style applied by the browser for accessibility and keyboard navigation.

> Sometimes image of NP shows the body but not the face of the NP (example: https://hollowverse.com/Jack_Black)

We should use `smartcrop.js` to crop the images to 1:1 with face recognition enabled.

> Nav bar scrolls until out of view then becomes sticky and switches to search bar

What should happen instead?

> Input field gets highlighted when I click on the magnifying glass

I'm not sure what you mean. Is that happening on search page? Is the search input empty? If so, this is the intended behavior (input field is required and when empty, the browser highlights it and shows "Please fill out this field" for me using Chrome). Anyway, there are some issues with the navbar and search box. I'll need to do some refactoring here.

> Cover image doesn't look good on wider screens. It looks as a repeating pattern. I removed the blurring to check and I saw Steve Austin's repeating dome as a cover image: 

I'm going to prevent the default `background-repeat` behavior and make the image stretch (although I'm not sure it will look better). Any suggestions?

> Quotation symbol is too strong, it clutters up the page

I'm going to make the text color for the symbol lighter.

> "Other interesting profiles" section requires side scrolling on narrower displays. Is that what we want?

I thought you mentioned that was OK. I don't have a preference. We can shrink the width of each image so that it does not overflow.

> Cards don't match design. These cards are elevated and have rounded corners. I think the flat cards are cleaner (and are closer to Facebook's design)

Yeah, I started with the same styling as the design but noticed that it looks somewhat out of place when compared to the FB comments card and the editorial summary card (both have shadows, unlike the related people images):

screenshot_20171225_213437

Also, this is what Facebook suggestions look on mobile for me (notice the shadows and the wider gap):

screenshot_20171225_224324

Anyway, it's your call really.

> Summary text is too crammed (needs bigger line-height or letter-spacing perhaps?)

I'll fix that.

## Priority

> Ryan Dunn page is empty

It's a stub on the old Hollowverse website. What should we do?

> Michael Jackson's image is missing

It turns out that because the image was manually replaced, the import script didn't know how to attach the source (and we are currently requiring that each image has a `sourceUrl` for attribution). I think I'll end up implementing some kind of override support for images so that we can force the scraper to use a specific image instead of searching for one using Wikipedia's API.

> Sometimes the Facebook Comments component is stuck on "Loading" (Steps to reproduce: load homepage, search for "st", select Steve Austin, scroll to bottom quickly and tap on one of the other interesting profiles, then scroll to bottom quickly and tap on another interesting profile. I did that a few times and Facebook comments wasn't loading any more)

I noticed that. I'll delay the call to `MutationObserver.takeRecords` until the promise of parsing the comments resolves and see if that fixes the issue.

> Sources is always open

This is already in the PR tasks. I'm working on it.

> On https://hollowverse.com/Chris_Martin page, at the bottom there's a link with the text "his wife". There are missing spaces around that text.

I thought https://github.com/hollowverse/scraper/pull/26 fixed this. I'm pretty sure I've removed all calls to `trim()` in the scrape function. I'll look into that again.

wholesomedev:
>screenshot_20171229_141459
>
>Are you referring to this? This is the default "focus ring" style applied by the browser for accessibility and keyboard navigation.

I don't see a similar pattern on Facebook or other websites. Can we eliminate this behavior?

>> Nav bar scrolls until out of view then becomes sticky and switches to search bar
>
>What should happen instead?

The nav bar should always be fully visible on the screen. And should animate into search bar when the user starts scrolling. That's how it is in `/design`

>> Input field gets highlighted when I click on the magnifying glass
>
>I'm not sure what you mean. Is that happening on search page? Is the search input empty? If so, this is the intended behavior (input field is required and when empty, the browser highlights it and shows "Please fill out this field" for me using Chrome). Anyway, there are some issues with the navbar and search box. I'll need to do some refactoring here.

I mean the gray border here 
image

>> Cover image doesn't look good on wider screens. It looks as a repeating pattern. I removed the blurring to check and I saw Steve Austin's repeating dome as a cover image: 
>
>I'm going to prevent the default `background-repeat` behavior and make the image stretch (although I'm not sure it will look better). Any suggestions?

I don't really have any suggestions. I'd say see if you can make something that you can be happy with. If not, I can also try to give it a shot.

>> "Other interesting profiles" section requires side scrolling on narrower displays. Is that what we want?
>
>I thought you mentioned that was OK. I don't have a preference. We can shrink the width of each image so that it does not overflow.

I think it's okay to keep it this way.

One thing I just noticed on my iPhone though is that when you swipe right or left, there's no inertial scrolling. The tiles make an abrupt stop instead. We need to add...

```
-webkit-overflow-scrolling: touch;
```

...to the scrolling element to enable inertial scrolling on those tiles.

>> Cards don't match design. These cards are elevated and have rounded corners. I think the flat cards are cleaner (and are closer to Facebook's design)
>
>Yeah, I started with the same styling as the design but noticed that it looks somewhat out of place when compared to the FB comments card and the editorial summary card (both have shadows, unlike the related people images):

Actually, my bad on this one. I just checked the design again and what you have matches it. I don't like the elevation though (generally speaking, for all cards), but that's something we can revisit later on.

>> Ryan Dunn page is empty
>
>It's a stub on the old Hollowverse website. What should we do?

Can we show some text on these types of pages? Something like

> Share what you know about the religion and political views of Ryan Dunn in the comments below

And can we prevent them from showing up in the search results?

forabi:
> I just checked the design again and what you have matches it.

I think you confused the screenshot in my last comment with what I ended up doing. The screenshot above is before I made the changes you currently see on the deployed branch. The deployed version does not really match what we have in `hollowverse/design`.

wholesomedev:
@forabi Hmmm, the cards in the deployed version look very similar to `hollowverse/design` from what I can tell. The only difference that I can see is a teeny tiny bit of extra elevation on the deployed version?

forabi:
> Hmmm, the cards in the deployed version look very similar to hollowverse/design from what I can tell. The only difference that I can see is a teeny tiny bit of extra elevation on the deployed version?

This and the gap between the image boxes is larger in the deployed version.

wholesomedev:
@forabi I see. Either design is fine by me for now.

forabi:
Tracking the progress of the issues so far:


# Future enhancements

* [x] When I click on the back button on desktop, a gray border appears around the back button. That shouldn't happen (removed focus outline)
* [ ] Sometimes image of NP shows the body but not the face of the NP (example: https://hollowverse.com/Jack_Black)
* [x] Nav bar scrolls until out of view then becomes sticky and switches to search bar (fixed, still needs some refactoring to enable animations)
* [x] Search text is too small (mismatch with design) (I checked the design and the computed font size for the search box is 14px, same as the this PR)
* [x] Input field gets highlighted when I click on the magnifying glass (removed focus outline)
* [x] Cover image doesn't look good on wider screens. It looks as a repeating pattern. (made the image stretch for now)
* [x] Quotation symbol is too strong, it clutters up the page (tuned down color and font size)
* [x] "Other interesting profiles" section requires side scrolling on narrower displays. (Enabled inertial scrolling)
* [x] Cards don't match design. These cards are elevated and have rounded corners. I think the flat cards are cleaner (and are closer to Facebook's design)
* [x] Summary text is too crammed (needs bigger line-height or letter-spacing perhaps?) (Increase line height to match editorial summary)

# Priority

* [x] Ryan Dunn page is empty (Added pitch text when no editorial summary exists for NP)
* [x] Michael Jackson's image is missing (fixed by https://github.com/hollowverse/scraper/pull/27 https://github.com/hollowverse/scraper/pull/28)
* [x] Sometimes the Facebook Comments component is stuck on "Loading" (fix attempted)
* [x] Sources is always open (fixed)
* [x] On https://hollowverse.com/Chris_Martin page, at the bottom there's a link with the text "his wife". There are missing spaces around that text. (Partially fixed, I've tracked down the issue to a bug in TypeORM query builder https://github.com/typeorm/typeorm/issues/1397). In the meantime, I've added extra spaces around each text node even if not needed, double white spaces are collapsed automatically (default browser behavior).
* [x] Prevent stub pages from showing in search results (updated Algolia's index with filtered list of notable people)

forabi:
@wholesomedev I've yet to find the cause of Facebook comments component being stuck at "Loading...". Other than that and the polyfill for `IntersectionObserver`, I think this is ready for review.

package.json
```diff
@@ -5,31 +5,59 @@
...
+    },
+    {
+      "path": "./dist/client/**/*.html",
+      "maxSize": "3 kB"
+    },
+    {
+      "path": "./dist/client/**/*.svg",
+      "maxSize": "30 kB"
+    }
+  ],
```
wholesomedev, line 55: This is pretty neat for keeping bundle sizes in check, always.

One suggestion: wouldn't it be better to move these to a separate config file?

src/app/assets/algolia/search-by-algolia.svg
```diff
@@ -0,0 +1 @@
...1.501.256.335.631.502 1.125.502a3.23 3.23 0 0 0 1.595-.419v1.101a2.53 2.53 0 0 1-.722.285 4.356 4.356 0 0 1-.932.086v.002zm8.277-.107h-1.268V9.506c0-.458-.092-.8-.277-1.026-.184-.226-.477-.338-.878-.338-.53 0-.919.158-1.168.475-.249.317-.373.848-.373 1.593v2.949h-1.262V4.801h1.262v2.122c0 .34-.021.704-.064 1.09h.081a1.76 1.76 0 0 1 .717-.666c.306-.158.663-.236 1.072-.236 1.439 0 2.159.725 2.159 2.175v3.873l-.001-.001zm7.649-6.048c.741 0 1.319.269 1.732.806.414.537.62 1.291.62 2.261 0 .974-.209 1.732-.628 2.275-.419.542-1.001.814-1.746.814-.752 0-1.336-.27-1.751-.811h-.086l-.231.704h-.945V4.801h1.262v1.987l-.021.655-.032.553h.054c.401-.591.992-.886 1.772-.886zm-.328 1.031c-.508 0-.875.149-1.098.448-.224.299-.339.799-.346 1.501v.086c0 .723.115 1.247.344 1.571.229.324.603.486 1.123.486.448 0 .787-.177 1.018-.532.231-.354.346-.867.346-1.536 0-1.35-.462-2.025-1.386-2.025l-.001.001zm3.244-.924h1.375l1.209 3.368c.183.48.304.931.365 1.354h.043c.032-.197.091-.436.177-.717.086-.281.541-1.616 1.364-4.004h1.364l-2.541 6.73c-.462 1.235-1.232 1.853-2.31 1.853-.279 0-.551-.03-.816-.091v-.999c.19.043.406.064.65.064.609 0 1.037-.353 1.284-1.058l.22-.559-2.385-5.941h.001z" fill="#1D3657"/></g></svg>
```
wholesomedev: Hmmm, why so many Algolia assets?

src/app/components/App/App.tsx
```diff
@@ -1,24 +1,41 @@
...
+import { Route, Switch } from 'react-router';
 
 import './App.global.scss';
 import * as classes from './App.module.scss';
 
 import { LoadableNotablePerson } from 'pages/NotablePerson/LoadableNotablePerson';
+import { LoadableSearchResults } from 'pages/SearchResults/LoadableSearchResults';
+import { LoadableAbout } from 'pages/About/LoadableAbout';
+import { LoadablePrivacyPolicy } from 'pages/PrivacyPolicy/LoadablePrivacyPolicy';
+import { LoadableHome } from 'pages/Home/LoadableHome';
```
wholesomedev, line 13: Looks like a lot of things are going to be `Loadable`. Would it (make sense|be possible) to create an HOC or some utility function to which you can pass a component in order to make `Loadable`? If we can do that, maybe we can also get rid of those `Loadable` files and simply export a loadable component form the source file?

src/app/components/Button/Button.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const Button = ({ className, ...rest }: ButtonProps) => (
+  <button className={cc([classes.root, className])} {...rest} />
+);
+
+export type LinkButtonProps = LinkProps;
+
+export const LinkButton = ({ className, ...rest }: LinkButtonProps) => (
+  <Link className={cc([classes.root, className])} {...rest} />
+);
```
wholesomedev, line 16: Nice!

src/app/components/Collapsable/Collapsable.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <div className={classes.root}>
+        <input
+          className={classes.input}
+          type="checkbox"
+          id={id}
+          onChange={this.handleChange}
+          checked={isOpen}
+          aria-checked={Boolean(isOpen)}
+        />
+        <label className={classes.label} {...{ for: id }}>
```
wholesomedev: If the labeled HTML element is inside the `label` tag, do you still `for`? If so, isn't `for` renamed to `htmlFor` in JSX? And with regard to `id`, wouldn't it be more reliable to use a UUID method from `lodash` instead of `random`?

package.json
```diff
@@ -5,31 +5,59 @@
...
     "server/build": "tsc --project ./src",
     "build": "npm-run-all -s clean -p app/build server/build",
     "start": "NODE_ENV=production run-p server/main-server/start",
-    "dev": "NODE_ENV=development API_ENDPOINT=${API_ENDPOINT:-http://localhost:8080/graphql} run-p app/graphql/dev server/app-server/dev",
-    "clean": "rm -rf ./dist ./src/app/components/**/*.module.scss.d.ts && mkdir -p ./dist",
+    "dev": "NODE_ENV=development API_ENDPOINT=${API_ENDPOINT:-http://localhost:8080/graphql} npm-run-all -l -s clean -p app/graphql/dev server/app-server/dev",
+    "clean": "del-cli './src/app/**/*.module.scss.d.ts' './dist' && mkdir -p ./dist",
     "validate-filenames": "validate-filenames -c commonconfig.js",
     "lint-staged": "lint-staged"
   },
```
wholesomedev: Should we add a command to the scripts for running the app locally but against production APIs?

Also, would it make sense to add `docker-compose` commands here as well?

src/app/components/Collapsable/Collapsable.module.scss
```diff
@@ -0,0 +1,51 @@
...
+        transform: translateY(-10px);
+      }
+
+      100% {
+        opacity: 1;
+        transform: translateY(0);
+      }
+    }
+  }
+}
```
wholesomedev, line 54: Can we animate the opening and closing of Collapsable components? (Not just the fading in and out of its children?)

src/app/components/EditorialSummary/EditorialSummary.module.scss
```diff
@@ -0,0 +1,39 @@
...
+    border-radius: 50%;
+    margin: 0 5px;
+    padding: 1px;
+    display: inline-block;
+    &:hover {
+      text-decoration: none;
+      background-color: $color-blue-opacity;
+    }
+  }
+}
```
wholesomedev, line 39: I have found single-direction margins to restore a lot of sanity to CSS spacing. I'm not sure if we've discussed this before but here's an article explaining it: https://csswizardry.com/2012/06/single-direction-margin-declarations/

Should we follow it, what do you think?

src/app/components/EditorialSummary/EditorialSummary.tsx
```diff
@@ -0,0 +1,200 @@
...
+                  <li key={nodeId} id={sourceId}>
+                    <a href={sourceUrl}>{sourceTitle || prettifiedUrl}</a>
+                    {sourceTitle ? ` ${prettifiedUrl}` : null}
+                    <a
+                      className={classes.backLink}
+                      href={`#${nodeId}`}
+                      role="button"
+                      aria-label="Go back to reference"
+                    >
+                      ↩
```
wholesomedev, line 192: This guy looks ugly on Mac
image

src/app/components/EditorialSummary/EditorialSummary.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+type State = {
+  shouldShowSources: boolean;
+};
+
+export class EditorialSummary extends React.PureComponent<Props, State> {
+  references = new Map<Node, Source>();
+
+  state: State = {
+    shouldShowSources: __IS_SERVER__,
```
wholesomedev: If `shouldShowSources` starts as `true` on the server, how does it get rendered being collapsed on the client? 🤔

src/app/components/EditorialSummary/EditorialSummary.tsx
```diff
@@ -0,0 +1,200 @@
...
+      number,
+      sourceId: `source_${number}`,
+      nodeId: `ref_${number}`,
+      sourceUrl: sourceUrl,
+      sourceTitle,
+    });
+  }
+
+  findChildren(node, nodes).forEach(findRefs(nodes, map));
+};
```
wholesomedev, line 124: Really cool stuff going on in the couple of functions above. 

One issue though: it's not clear what set of problems this code is solving for. Could you add documentation for why you're doing the things you're doing? Like why are you collecting refs like this, why does it make sense to traverse nodes the way you're doing, why are you adding spaces around `a` tags, etc?

src/app/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,35 @@
...
+  <footer className={classes.root}>
+    <ul className={classes.list}>
+      <li>
+        <Link to="/about">About</Link>
+      </li>
+      <li>
+        <Link to="/privacy-policy">Privacy</Link>
+      </li>
+      <li>
+        <a href="https://twitter.com/hollowverse">Twitter</a>
```
wholesomedev: Can we remove this link? This account is not active anymore.

src/app/components/LazyImage/LazyImage.tsx
```diff
@@ -0,0 +1,39 @@
...
+            );
+          }
+
+          if (!inView) {
+            return this.props.loadingComponent;
+          }
+
+          return (
+            <div>
+              <Image {...imageProps} />
```
wholesomedev: Why wrap `Image` in a `div` here? Why not just return `Image`?

src/app/components/LazyImage/LazyImage.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Image, Props as ImageProps } from 'components/Image/Image';
+
+type Props = ImageProps & Pick<IntersectionObserverProps, 'rootMargin'>;
+
+export class LazyImage extends React.PureComponent<Props> {
+  render() {
+    const { rootMargin, ...imageProps } = this.props;
+
+    return (
+      <OptionalIntersectionObserver rootMargin="25%" triggerOnce>
```
wholesomedev: According to IntersectionObserver documentation, you can use `rootMargin` as a percentage only if you specify a `root` element: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API

I don't see us specifying a `root` element?

src/app/components/NavBar/NavBar.tsx
```diff
@@ -1,8 +1,134 @@
...
+import { RouteComponentProps, Route, Switch } from 'react-router';
+
 import * as classes from './NavBar.module.scss';
 
-export const NavBar = (props: { title: string }) => (
-  <div className={classes.navbar}>
-    <a href="/">{props.title}</a>
-  </div>
-);
+import { Sticky } from 'components/Sticky/Sticky';
```
wholesomedev, line 8: I feel that at this point, it might be better to eliminate the `Sticky` component. It's not doing much and is kind of misnamed.

src/app/components/NavBar/NavBar.tsx
```diff
@@ -1,8 +1,134 @@
...
+  isSearchInProgress: boolean;
+  shouldFocusSearch: boolean;
+  isSearchPage: boolean;
+  isHomePage: boolean;
+};
+
+export type DispatchProps = {
+  goToSearch(): any;
+  searchQueryChanged({ query }: { query: string }): any;
+  setShouldFocusSearch(isFocused: boolean): any;
```
wholesomedev: Hmm, can't you tell TypeScript that these functions are the same as what you have in `search/actions.ts`?

src/app/components/NavBar/NavBar.tsx
```diff
@@ -1,8 +1,134 @@
...
+};
+
+type Props = OwnProps & StateProps & DispatchProps;
+
+export const NavBar = class extends React.Component<
+  Props & RouteComponentProps<any>
+> {
+  goBack = (e: React.MouseEvent<HTMLElement>) => {
+    e.preventDefault();
```
wholesomedev: Why is `preventDefault` required here?

src/app/components/NavBar/NavBar.tsx
```diff
@@ -1,8 +1,134 @@
...
+                ])}
+              >
+                <SvgIcon size={20} {...backIcon} />
+                <span className="sr-only">Go Back</span>
+              </NavBarButton>,
+              <div className={classes.view} key="view">
+                <Switch>
+                  <Route path="/search">{searchView}</Route>
+                  <Route>
+                    {() => {
```
wholesomedev: Isn't this better written as...

```jsx
{isSticking || shouldFocusSearch ? searchView : <div className={classes.logoViewInner}>...etc</div>}
```

...without the anonymous function?

src/app/components/NavBar/NavBar.tsx
```diff
@@ -1,8 +1,134 @@
...
+                          </div>
+                          <NavBarLink
+                            className={cc([
+                              classes.button,
+                              { [classes.isHidden]: isHomePage },
+                            ])}
+                            to="/search"
+                          >
+                            <SvgIcon size={20} {...searchIcon} />
+                            <span className="sr-only">Search</span>
```
wholesomedev: These are nice. My concern is, how do make sure that this work is not in vain? How do we make sure that Hollowverse provides a sufficient level of accessibility? That we are not wasting our time with these while our website remains largely inaccessible? Are there linting rules or best practices or some kind of tools or approaches that can guarantee that our website has a sufficient level of accessibility? Should we make it easy for users to report accessibility issues? (Just some thoughts)

src/app/components/NavBar/NavBarButton.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  color: inherit;
+  display: flex !important;
+  align-items: center;
+  align-self: stretch;
+  background: transparent;
+  transition: opacity 0.1s ease-in-out;
+  border: 0 !important;
+  flex-grow: 0;
+  padding: 8px 15px !important;
+  background: transparent !important;
```
wholesomedev, line 13: Why is the `!important` needed on some of these rules?

forabi:
The deployed version on the `new-app` branch has a potential fix for the Facebook comments component issue. Can you confirm that the issue is fixed?

src/app/components/NavBar/SearchView.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { Route, Switch } from 'react-router';
+
+type Props = {
+  inputValue?: string;
+  isFocused?: boolean;
+  shouldFocusSearch: boolean;
+  isSearchPage: boolean;
+  goToSearch(): any;
+  setShouldFocusSearch(isFocused: boolean): any;
+  searchQueryChanged({ query }: { query: string }): any;
```
wholesomedev: These are duplicated. Is it possible to type these once in the Redux world and import their types where needed?

...or maybe connect the `SearchView` separately and import it where needed?

src/app/components/NavBar/SearchView.tsx
```diff
@@ -0,0 +1,102 @@
...
+    this.focusIfNecessary(nextProps);
+  }
+
+  componentDidMount() {
+    this.focusIfNecessary();
+  }
+
+  focusIfNecessary = (props: Props = this.props) => {
+    if (props.isFocused && this.searchInput) {
+      this.searchInput.focus();
```
wholesomedev: Hmm, maybe React should provide the ability to focus a component as a prop...

src/app/components/NavBar/SearchView.tsx
```diff
@@ -0,0 +1,102 @@
...
+              <span className="sr-only">Search</span>
+            </NavBarButton>
+          </Route>
+          <Route>
+            <NavBarLink className={classes.button} to="/search">
+              <SvgIcon size={20} {...searchIcon} />
+              <span className="sr-only">Search</span>
+            </NavBarLink>
+          </Route>
+        </Switch>
```
wholesomedev, line 107: It's interesting to me to see React Router being used in this way. I think it's cool. But is it meant to be used this way? All you really need here is to know the route where the user currently is, right?

src/app/components/PersonDetails/PersonDetails.module.scss
```diff
@@ -1,51 +1,89 @@
 @import 'variables.scss';
 
-.person-details {
+.root {
+  .cover-photo {
+    visibility: hidden;
+  }
+}
+
+@supports (filter: blur(5px)) {
```
wholesomedev, line 9: Hmmm, how does Sass know whether a browser supports certain CSS rules at runtime? 🤔

src/app/components/PersonDetails/PersonDetails.tsx
```diff
@@ -2,45 +2,64 @@ import * as React from 'react';
...
-        </li>
-      ))}
-    </ul>
-    {summary && (
-      <div className={classes.personDetailsAbout}>
-        {summary.split('\n').map((paragraph, i) => <p key={i}>{paragraph}</p>)}
-      </div>
-    )}
+    <div className={classes.content}>
+      {photo ? (
```
wholesomedev: I'm just curious why you prefer `condition ? <Component/> : null` over just `condition && <Component/>`?

src/app/components/Post/Post.tsx
```diff
@@ -0,0 +1,86 @@
...
+  sourceName?: string | null;
+  sourceUrl?: string | null;
+  organizationName?: string | null;
+  organizationWebsiteUrl?: string | null;
+  labels?: Array<{
+    id: string;
+    text: string;
+  }>;
+  children?: React.ReactNode;
+};
```
wholesomedev, line 24: I feel like a lot of these things we've already typed in the APIs, can't we just reuse that?

(I see that this Post component is not being used any where as of now)

src/app/components/Square/Square.tsx
```diff
@@ -0,0 +1,17 @@
...
+ */
+export const Square = ({
+  className,
+  children,
+  ...rest
+}: React.HTMLAttributes<HTMLDivElement>) => (
+  <div className={cc([classes.root, className])} {...rest}>
+    <div className={classes.content}>{children}</div>
+  </div>
+);
```
wholesomedev, line 17: Never occurred to me that this is hard to do in responsive CSS! 👍

src/app/components/Status/Status.tsx
```diff
@@ -0,0 +1,57 @@
...
+      if (isRedirection(this.props)) {
+        this.props.setRedirectionUrl(this.props.redirectTo);
+      }
+    }
+
+    render() {
+      return null;
+    }
+  },
+);
```
wholesomedev, line 66: Wait...redirections, status codes....in a front-end framework, React...I'm utterly confused...what's going on here?

src/app/components/TextPage/TextPage.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import cc from 'classcat';
+
+import * as classes from './TextPage.module.scss';
+
+export const TextPage = ({
+  className,
+  ...rest
+}: React.HTMLAttributes<HTMLDivElement>) => (
+  <div className={cc([className, classes.root])} {...rest} />
```
wholesomedev, line 10: I like how you're passing `children` here. It's never occurred to me to pass `children` as a prop like this 👍

src/app/hocs/AsyncComponent/AsyncComponent.tsx
```diff
@@ -1,38 +1,65 @@
...
    */
   timeout?: number | null;
 
+  /**
+   * Time in milliseconds to wait before the component is considered
+   * to have started loading. The `load()` function will still be called
+   * as soon as the component mounts. This is only intended to avoid flashing
+   * of loading component in image components when using server side rendering
+   * and the image is already cached client-side.
+   * In such cases, the loading component would should for a brief moment before
```
wholesomedev: typo: "should" => "flash"

But you don't mention how adding this delay prevents the flashing? Or does it only delay the flashing?

src/app/hocs/AsyncComponent/AsyncComponent.tsx
```diff
@@ -41,47 +68,61 @@ type State<T> = {
...
-        ];
-
-        const { timeout = null } = this.props;
-        if (timeout !== null) {
+        const promises: Array<Promise<Partial<State<T | null>>>> = [];
+
+        promises.push(
+          loadPromise.then(value => ({
+            value,
+            isInProgress: false as false,
```
wholesomedev: What does `false as false` accomplish here?

src/app/hocs/WithData/WithData.tsx
```diff
@@ -0,0 +1,142 @@
...
+  StateProps,
+  DispatchProps,
+  OwnProps,
+  StoreState
+>(
+  (state, ownProps) => ({
+    result: getResolvedDataForKey(state)(ownProps.dataKey),
+  }),
+  { requestData },
+)(Wrapper);
```
wholesomedev, line 142: 👍

src/app/pages/About/About.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <h1>About</h1>
+    <p>
+      It is often said that one should avoid discussing politics or religion in
+      polite company. But these are some of the most interesting topics and I
+      hope I can get you talking about them!
+    </p>
+    <p>
+      Email me at <a href="mailto:msafi@msafi.com">msafi@msafi.com</a> if you’d
+      like to discuss anything.
+    </p>
```
wholesomedev: Can this paragraph...

```jsx
<p>
      Email me at <a href="mailto:msafi@msafi.com">msafi@msafi.com</a> if you’d
      like to discuss anything.
</p>
```

...be changed to...

```jsx
<p>
      Email us at <a href="mailto:hello@hollowverse.com">hello@hollowverse.com</a> if you’d
      like to discuss anything.
</p>
```

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -1,210 +1,171 @@
...
+                    </Card>
+                  ) : (
+                    <div className={classes.stub}>
+                      Share what you know about the religion and political views
+                      of {name} in the comments below
+                    </div>
+                  )}
+                </article>
+                {notablePerson.relatedPeople.length ? (
+                  <div>
```
wholesomedev: This component is growing too big. Should the `relatedPeople` functionality be spun off into its own component?

src/app/pages/NotablePerson/NotablePersonSkeleton.module.scss
```diff
@@ -0,0 +1,111 @@
...
+
+@keyframes animate-gradient {
+  0% {
+    transform: translateX(-180px);
+  }
+
+  100% {
+    transform: translateX(780px);
+  }
+}
```
wholesomedev, line 111: This approach of making skeleton components requires more than a small amount of effort. The CSS of these components has to very closely match that of the fleshed component. Otherwise, skeleton components don't work as well.

I have an idea about creating skeleton components that might make this both easier and work better. I haven't put this idea to the test yet, but I think it might work.

What if instead of rendering a skeleton component while you wait for data, you render your original component but with _dummy data_ and _additional CSS_ to format the dummy data in a certain way to indicate that it's progressing, like render blurry text with animated gradient or a blurry photo with animated gradient. Regardless of how you style the dummy data, don't you think this approach would work?

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -0,0 +1,164 @@
...
+                  __IS_SERVER__ &&
+                  value.hits.length === 1
+                ) {
+                  return (
+                    <Status code={301} redirectTo={`${value.hits[0].slug}`} />
+                  );
+                }
+
+                return (
+                  <div>
```
wholesomedev, line 123: I think breaking up this large render function either into multiple components or separate render methods would improve readability and comprehension here.

wholesomedev (changes requested):
image

😄

I still need to review Redux stuff, review other files, and leave some more comments about UI related issues. So...

...to be continued...

package.json
forabi: You can do:

```
API_ENDPOINT=https://api.hollowverse.com/graphql yarn dev
```

We can add shortcuts but I'm running out of meaningful names for the commands 😅

forabi: Regarding `docker-compose`, `docker-compose up` is all we have to remember. It calls `yarn dev` in the Docker container.

package.json
forabi: `bundlesize` does not currently support a separate configuration file (https://github.com/siddharthkp/bundlesize/issues/162)

src/app/components/Collapsable/Collapsable.tsx
forabi: The input, label, and children container all have to be siblings in order to hide the children with pure CSS using the sibling selector:

```scss
.input:not(:checked) ~ .children {
  display: none;
}
```

If we wrap the `input` in a `label`, we can no longer target the children and we'll need to use JS for that. We'll also need to replicate some of the browser behavior for accessibility.

> If so, isn't for renamed to htmlFor in JSX

Correct. Unfortunately, Preact prefers standard attribute names, e.g. `class`, `for`. `preact-compat` should convert `className` and `htmlFor` to their Preact equivalents and indeed it does so. However, I noticed that `htmlFor` only works for the browser build. There seems to be an issue in `preact/server` where the `htmlFor` attribute is passed as-is because when I disable JS, the generated server markup has `htmlfor` instead of `for` and the collapsable component does not toggle when I click on the label.

The `{...{ for: id }}` tricks TypeScript into ignoring the "unknown" property `for`, since TS expects `htmlFor` instead.

> wouldn't it be more reliable to use a UUID method from `lodash` instead of `random`

Lodash does not have a `uuid` function as far as I know. It _does_ have `uniqueId` but that just returns `i + 1` every time it's called. I started with `uniqueId` but since it's also called on the server, the number may grow indefinitely until the Node process restarts.

src/app/components/EditorialSummary/EditorialSummary.tsx
forabi: The server markup will have the sources shown by default (`shouldShowSources: true`). When JS is executed, Preact calls `render` and sees that, in the browser build, `shouldShowSources` is now `false`, so it hides them.

src/app/components/LazyImage/LazyImage.tsx
forabi: This is for styling consistency with `<noscript><Image /></noscript>` so that we do not assume that `Image` is a direct child.

src/app/components/NavBar/NavBar.tsx
forabi: Actually it's not required anymore. I was using a `Link` previously. I'm going to remove it.

src/app/components/NavBar/NavBar.tsx
forabi: I'm following some of the basic accessibility recommendations provided by Lighthouse. TSLint also enforces some accessibility rules.

Generally speaking, we should at least check:

* Keyboard navigation
* Focusable elements have some visual indication
* Buttons, images and links are correctly labeled.
* Menus are correctly formatted (`ul` and `ol` have direct `li` children)
* Use of semantic markup. `nav`, `main`, `article`, headings... etc. This greatly helps screen readers.
* Use built-in browser components, like inputs and buttons, instead of replicating their behavior with JS/CSS. Most of the browser components have behaviors that are either hard or impossible to replicate with JS/CSS.
* Testing with screen readers/OS accessibility services.

Unfortunately Linux accessibility services are not great. I used to use ChromeVox as a screen reader for Chrome but it broke for some reason.

I tested some parts of the UI with Android's accessibility service and once I get my hands on an iPhone I'll do some further testing.

I've been following Rob Dodson's [a11ycasts](https://www.youtube.com/watch?v=V1yoOLhx_qA&list=PLNYkxOF6rcICWx0C9LVWWVqvHlYJyqw7g) for a while and it's great!

src/app/components/NavBar/NavBarButton.module.scss
forabi: `NavBarButton` extends `Button`, these are needed to override the default `Button` styles. I think there's room for improvement with regard to CSS.

src/app/components/NavBar/SearchView.tsx
forabi: Yeah, connecting `SearchView` is probably better.

src/app/components/NavBar/SearchView.tsx
forabi: HTML has `autoFocus` on `input` but it has issues with React rendering.

src/app/components/PersonDetails/PersonDetails.module.scss
forabi: `@supports` is native CSS, not Sass. Support is checked at runtime.

src/app/hocs/AsyncComponent/AsyncComponent.tsx
forabi: TypeScript is looking at this object and trying to infer its type, its considers `isInProgress` to be a `boolean`. It then tries to check if that type is compatible with any of the subtypes of `AsyncResult`. However, since `AsyncResult` types are very strict regarding the value of `isInProgress`, TypeScript complains that `isInProgress` as a `boolean` does not match any of the combinations of `isInProgress`, `value` and `hasError` of `AsyncResult`.

src/app/components/Status/Status.tsx
forabi: These are used for SSR (https://github.com/forabi/hollowverse/blob/92729b6fbb24ae8a4f3c20d1b0b1e2b5fa70a7fc/src/app/server.tsx#L63). Usage is here https://github.com/forabi/hollowverse/blob/92729b6fbb24ae8a4f3c20d1b0b1e2b5fa70a7fc/src/app/pages/NotablePerson/NotablePerson.tsx#L83

src/app/hocs/AsyncComponent/AsyncComponent.tsx
forabi: Anyway, it might be better to use a type assertion (`as Partial<SuccessResult<T>>`) instead.

src/app/components/PersonDetails/PersonDetails.tsx
forabi: I remember somebody on Twitter mentioned that `condition && <Component />` has a caveat. If `condition` is `true` and `<Component />` returns a falsey value, the whole expression evaluates to `false` and it might be coerced to a string (`"false"`) in some cases.

src/app/components/Post/Post.tsx
forabi: I generally prefer to keep components completely unaware of the data layer. It makes them easier to test in the future. I like to think that the component props just _happen to_ match the API types.

src/app/components/NavBar/SearchView.tsx
forabi: The React Router guys have been pushing for the idea that `Route`s are just components that render when a condition matches and there's nothing magical about them. Actually, `Route` can be simply replaced with an `if` statement (provided that we can access the `location` prop somehow).

> All you really need here is to know the route where the user currently is, right?

I don't see how that could be done if not using React Router?

src/app/components/Collapsable/Collapsable.module.scss
forabi: The icon rotates when the `Collapsable` state changes. Any other suggestions? I feel we are pushing the limits of CSS animations and we might need to bring in some animation library in the future. CSS animations have the benefit of working without JS though (and they usually perform better too).

src/app/components/NavBar/NavBar.tsx
forabi: When using the function version, the JSX expressions are only evaluated when the route matches. So this could be better performance-wise. It's not the reason why I used an anonymous function though. I may have forgotten to remove it during refactoring.

What do you think?

src/app/components/App/App.tsx
forabi: AThat would not work.

To get Webpack to code split and create a chunk for the loadable component, we need to use the _dynamic_  `import()` with the path to the component. If we import the component and pass it to a HOC to make it loadable... well, it's not going to be in another chunk, because we just imported it using the standard `import { Component } from 'path/to/Component'`.


It wouldn't work with a HOC that accepts the _path_ to the file either. The dynamic `import()` call must be statically analyzable so that Webpack knows what files to include in the dynamic chunks. If we create a HOC for loadable components, this is what `import()` call could look like:

```typescript
import(props.pathToComponent)
```

Webpack will have no idea where to find the files at build time. It's actually pretty smart and _can_ create chunks if the string is partially static:

```typescript
import(`components/${componentName}`);
```

However, in that case, Webpack will go ahead and create chunks for every single file that is found in `components`. We can limit the files pattern using Webpack's [magic comments](https://medium.com/faceyspacey/how-to-use-webpacks-new-magic-comment-feature-with-react-universal-component-ssr-a38fd3e296a) but it feels a little bit hacky to me. I don't like writing build tool-specific comments in code.

src/app/components/Footer/Footer.tsx
forabi: Removed.

src/app/assets/algolia/search-by-algolia.svg
forabi: I just removed the unused ones.

src/app/components/EditorialSummary/EditorialSummary.module.scss
forabi: Sounds good! I'll see where we can apply that.

src/app/components/LazyImage/LazyImage.tsx
forabi: Interesting. It's weird that it's actually working. You can confirm by looking at the Network tab and checking that the requests for the related people images start as soon as you scroll down to the collapsed sources label. Could `react-intersection-observer` be doing something to make that work by default?

src/app/components/LazyImage/LazyImage.tsx
forabi: This has been updated since you reviewed it. It's no longer wrapped in a `div`.

src/app/components/NavBar/NavBar.tsx
forabi: See https://github.com/hollowverse/hollowverse/pull/225#discussion_r159154766

> I generally prefer to keep components completely unaware of the data layer. It makes them easier to test in the future. I like to think that the component props just happen to match the API types.

Same applies for Redux props.

src/app/pages/About/About.tsx
forabi: Done

src/app/pages/NotablePerson/NotablePersonSkeleton.module.scss
forabi: I thought about that before going with this. The problem is that sometimes we need to show the skeleton component in place of the actual component while that component is loading, like the `NotablePerson` component above. We cannot use the component while it's loading obviously. So the approach you suggested only works when we have the component loaded but we are loading the data it needs to display.

forabi: By the way, the approach you described is exactly what https://github.com/lightsinthesky/react-skeleton does

package.json
wholesomedev: how about `yarn dev-prod-api`? I know it's verbose, but having the command in `package.json` helps with memory for someone like me who's not as intimately involved with the setup as you are.

src/app/components/Collapsable/Collapsable.module.scss
wholesomedev: I was going to suggest animating between `display: none` and `display: inline-flex` or animating between `height: 0` and `height: auto`, but a quick search revealed that that's not easily possible...so I'm not sure...

src/app/hocs/AsyncComponent/AsyncComponent.tsx
forabi: I've expanded the comment to explain that in more detail. Is it more clear now?

> If the delay is set to, say, 200ms, the browser has 200ms to show the image before we trigger the "loading" state. If the image is already in the browser cache, the browser will usually finish loading the image before the 200ms deadline, and the loading state will never trigger, thus we avoid flashing.

src/app/components/Collapsable/Collapsable.tsx
wholesomedev: I see. Should this be documented?

With regard to UUID, I've used [this](https://stackoverflow.com/a/8809472/604296) in the past. Or we can continue to use `random`. No objections from me.

src/app/components/EditorialSummary/EditorialSummary.tsx
wholesomedev: But there's probably some styling that we can apply that will prevent Mac from rendering an emoji like the above...I say that because in the code, I see plain arrow, only when it is rendered on the screen do I see the emoji arrow. That leads me to believe there's certain styling/font perhaps that keeps the plain arrow?

src/app/components/NavBar/NavBar.tsx
wholesomedev: I think the anonymous function makes it seems as though there's something special going on here, which isn't the case.

src/app/components/PersonDetails/PersonDetails.module.scss
wholesomedev: TIL!

src/app/components/NavBar/SearchView.tsx
wholesomedev: It's just a bit hard to reason about when encountered for the first time. I found myself staring at the code for a good while until I finally got what's going on. With that said, I'm actually okay with it (as long as it's not a non-standard practice). It's just that I'm unfamiliar with the pattern.

> provided that we can access the `location` prop somehow

I think `react-router-redux` puts the router state in Redux. That way you can connect the component to the store and get the `location` prop from there.

src/app/components/Collapsable/Collapsable.module.scss
forabi: Not to mention that animating [`height` will trigger layout every single frame](https://csstriggers.com/height). Unlike things like `opacity` and `transform`, `height` is not one of the cheap properties that can be offloaded to the GPU.

src/app/components/Status/Status.tsx
wholesomedev: Aha, the `Status` component renders nothing, and is only used for its side effect on the Redux store (which is then queried on the server-side). I was going to ask "so why don't you just use plain Redux actions", but I like how this component can be conveniently embedded within the JSX logic...

But...come to think of it more...you could also embed a function a function in the JSX logic...hmmm 🤔

src/app/components/NavBar/SearchView.tsx
forabi: > I think react-router-redux puts the router state in Redux. That way you can connect the component to the store and get the location prop from there.

I was just stating that we need the `location` prop for this. React Router exports a HOC, `withRouter`, that provides `location`.

src/app/pages/NotablePerson/NotablePersonSkeleton.module.scss
wholesomedev: Ah, interesting. Well, it doesn't have to be part of this PR but it's something we can continue to think about.

>The problem is that sometimes we need to show the skeleton component in place of the actual component while that component is loading, like the `NotablePerson` component above.

One way to solve this problem is to create `NotablePersonBody` component that can be rendered differently inside the `NotablePerson` component depending on the availability of data.

package.json
forabi: Do you think it makes sense to default to `API_ENDPOINT` being the production API instead of the local instance? If so, we need to document that CORS has to be disabled during development.

src/app/components/Status/Status.tsx
forabi: Yeah, but I've not seen many functions called in `render`. Actually this `Status` component is inspired by this https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/guides/server-rendering.md#404-401-or-any-other-status

src/app/components/NavBar/NavBar.tsx
forabi: Why do you think it's named? It _is_ responsible for applying the sticky styles depending on the intersection state changes.

forabi: misnamed*

src/app/pages/SearchResults/SearchResults.tsx
forabi: Done

src/app/components/NavBar/SearchView.tsx
forabi: Done.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: Done.

src/app/components/NavBar/NavBar.tsx
forabi: Removed the anonymous function

src/app/components/NavBar/NavBar.tsx
wholesomedev: Okay, I find it hard to tell what `Sticky` is doing...

It's a `position: relative` element. Inside of it, there's a `position: fixed` element, which is _always sticky_. But the children of this, always sticky, element receive a flag called `isSticky`, which is either `true` or `false` based whether the `position: relative` container is in view?

It seems like the `Sticky` component is trying to relay information between two elements where one element is interested in whether another element is in view or not. So it's not really about stickiness...

src/app/components/Collapsable/Collapsable.tsx
forabi: I've replaced `random` with the suggested UUID function. Actually, the previous code had a problem: `random()` always returned either `0` or `1` (looks like those are the default parameters of the function)

package.json
wholesomedev: I've been using 

```
API_ENDPOINT=https://api.hollowverse.com/graphql yarn dev
```

in my review of this PR.

There's an issue where every time I navigate between NP pages, the app breaks due to some issue with CORS preflight check. Then I click on the "Reload" button and the page loads. I'm not sure why that's happening. But if we can fix that, then I think it's fine to default `yarn dev` to run against production API.

forabi: That means that the browser extension you are using is not doing anything. When you reload the page, what you is the server-rendered markup of the page. It has all the data because the server can freely access the API. When you navigate to a new page, the call to the API now happens in the browser and it fails because of CORS restrictions. You may need to add `https://api.hollowverse.com/graphql` to your extension exception list and keep DevTools open while testing (looks like it only works with DevTools open, not sure).

src/app/components/NavBar/NavBar.tsx
forabi: > there's a position: fixed element, which is always sticky

Correct. But you forgot about the placeholder element, which is actually responsible for maintaining the original element's space while the element is sticking. So what `Sticky` is doing is:

1. Make the target element `fixed`.
2. Create a placeholder element to maintain the space of the target element.
3. Now the two elements are laid in the same space. The target element is on top because `fixed` creates a new stacking context.
4. Now when the user scrolls, the placeholder will go out of view, while the target will stay `fixed` at the top of the viewport. We monitor the placeholder element so that when it's not in view, it means that the target element has become effectively sticky from the user's point of view.
5. We communicate the state changes in the render prop in case we want to do something with that information, which we do in this case -- we want to transform the navbar to `SearchView`.

src/app/components/EditorialSummary/EditorialSummary.tsx
forabi: This is relevent: https://stackoverflow.com/questions/32915485/how-to-prevent-unicode-characters-from-rendering-as-emoji-in-html-from-javascrip

> Update: It seems (from comments below) there is a text presentation selector, FE0E, available to enforce text-vs-emoji. The selector is concatenated as a suffix without space onto the character's code, such as &#x231b;&#xFE0E for HTML hex or \u231B\uFE0E for JS. However, it is simply not honored by all browsers (eg Chrome and Edge).

src/app/store/features/analytics/epic.ts
```diff
@@ -0,0 +1,38 @@
...
+import 'rxjs/add/operator/ignoreElements';
+
+import { importGlobalScript } from 'helpers/importGlobalScript';
+import { once } from 'lodash';
+
+const GA_URL = 'https://www.google-analytics.com/analytics.js';
+
+const initScript = once(async () => {
+  await importGlobalScript(GA_URL);
+  ga('create', 'UA-110141722-1', 'auto');
```
wholesomedev, line 18: Stupid question: how does TypeScript know that this got added to the global namespace?

src/app/store/features/data/actions.ts
```diff
@@ -0,0 +1,4 @@
+import { createActionCreator } from 'store/helpers';
```
wholesomedev: I see how `features/data` is being used with the `WithData` component and it makes sense...but everything is `data`, so can we differentiate this with a more descriptive name? Or at least add documentation or a README.md file in this folder?

src/app/store/features/search/updateUrlEpic.ts
```diff
@@ -0,0 +1,50 @@
...
+    .map(action => {
+      const { query } = (action as Action<'SEARCH_QUERY_CHANGED'>).payload;
+      const searchParams = new URLSearchParams();
+      searchParams.append('query', query);
+      const descriptor = {
+        pathname: '/search',
+        search: searchParams.toString(),
+      };
+
+      const location = store.getState().routing.location;
```
wholesomedev: Does it matter if you use an existing selector here or not?

src/app/store/helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+  GenericAction,
+} from 'store/types';
+
+/**
+ * A helper functions that creates an action of a specific type,
+ * the payload must correspond to the action type.
+ */
+export function createAction<T extends ActionType>(
+  type: T,
+  payload: TypeToPayload[T],
```
wholesomedev: Man, naming things _is_ hard. This should really be called `ReduxActionTypeToPayloadTypeScriptType` but naming it that would probably cause a riot 😂

src/app/store/helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+): Action<T> {
+  return {
+    type,
+    payload,
+  };
+}
+
+/** A better typed version of `redux-action`'s `createAction`.
+ * It only accepts valid action types and returs an action creator
```
wholesomedev: `returs` => `returns`

src/app/store/helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (reducer !== undefined) {
+      return reducer(state, action);
+    }
+    if (state !== undefined) {
+      return state;
+    }
+
+    return defaultState;
+  };
+}
```
wholesomedev, line 58: 👍

wholesomedev (approved):
Really like the Redux typing, selectors, and observables. Awesome stuff! 👍

package.json
wholesomedev: Okay, so now I tried this extension instead and it works fine: https://chrome.google.com/webstore/detail/cors-toggle/jioikioepegflmdnbocfhgmpmopmjkim/related?hl=en

So I'm okay with making `yarn dev` work against prod API by default

wholesomedev:
I have a few comments about UI, but I don't think these comments should block this PR. This PR has grown too big and I think we should merge it ASAP and continue work in smaller chunks:

* I think we need to increase the size of the search box text
* "Jennifer Love Hewitt" card is bigger than other NPs on Jerry Lewis's page
* Margin around FB comments is too big
* In `hollowverse/design` the cover image has default highlights on the side, we're missing that here (I think we still haven't gotten the cover image right. I think we need to work more on it)
* On iPhone, when user focuses an input field (e.g. search box or Facebook comments box), the page zooms in for no reason
* When CORS is disabled, and I click on the "Reload" page button, the page registers two browser history items when it loads, so I can't back out of it without pressing the back button twice
* Clicks outside of the NP image redirect to Wikipedia. I think only clicks inside the NP's image should redirect to Wikipedia
* Issues on Ashanti (singer) page:
  * The search icon doesn't show on a hard reload
  * The cover image doesn't show. If I navigate from Ashanti to Keyshia Cole's page and then back to Ashanti, I see Keyshia's cover image on Ashanti's page

forabi:
I fixed some of these issues. Let's merge this and continue iterating in future PRs.

* [x] I think we need to increase the size of the search box text
* [x] "Jennifer Love Hewitt" card is bigger than other NPs on Jerry Lewis's page
* [x] Margin around FB comments is too big
* [ ] In `hollowverse/design` the cover image has default highlights on the side, we're missing that here (I think we still haven't gotten the cover image right. I think we need to work more on it)
* [ ] On iPhone, when user focuses an input field (e.g. search box or Facebook comments box), the page zooms in for no reason
* [ ] When CORS is disabled, and I click on the "Reload" page button, the page registers two browser history items when it loads, so I can't back out of it without pressing the back button twice
* [x] Clicks outside of the NP image redirect to Wikipedia. I think only clicks inside the NP's image should redirect to Wikipedia
* [ ] Issues on Ashanti (singer) page:
  * [ ] The search icon doesn't show on a hard reload
  * [x] The cover image doesn't show. If I navigate from Ashanti to Keyshia Cole's page and then back to Ashanti, I see Keyshia's cover image on Ashanti's page

src/app/store/features/search/updateUrlEpic.ts
forabi: Updated

src/app/store/helpers.ts
forabi: Fixed

src/app/store/features/analytics/epic.ts
forabi: TypeScript has a concept of "module augmentation". Basically it allows modules to extend the prototype of another module.

I assume RxJS imports are written this way:

```typescript
// module 'rxjs/add/map'

import { Observable } from 'rxjs/Observable';

Observable.prototype.map = /** ... */

```

However, TypeScript has an issue where if you import one operator in File A, it will assume that every other file now can access that operator on `Observable`. That's not always correct because we could try to access the operator in File B before we import File A. It's not aware of the order of dependencies and has no idea that we are doing things like code splitting.

That's where `rxjs-tslint-rules` help. The rules enforce that every operator is imported in the same module it's used in and that there are no unused imports.

src/app/store/features/data/actions.ts
forabi: I changed the name to `asyncData` and added a README.md to document how this feature works with `WithData`.

forabi:
I've changed the default `API_ENDPOINT` to the production API and updated the wiki to include a note on CORS issue.

PR #230, opened 2018-01-05
Launch new Tom Hanks page

### LINKED ISSUE #248: Fix browser support list
We've changed browser support list recently and we might need to make further modification to support a higher percentage of our users

forabi:
Tasks:
* [x] Log page load success/failure
* [x] Log page redirections
* [x] Log unhandled JavaScript errors (should reference the source position of the error)
* [x] Log search result clicked
* [x] Log search page visited
* [x] Batch and schedule events to be processed when the browser is idle (if possible)
* [x] Remove note about CORS issues in wiki

Closes #248 #237 #243 #242 #234


src/app/components/Collapsable/Collapsable.tsx
```diff
@@ -5,7 +5,7 @@ import { SvgIcon } from 'components/SvgIcon/SvgIcon';
 
 import dropdownIcon from 'assets/iconDropdown.svg';
 
-import { generateUuid } from 'helpers/generateUuid';
+import { generateRandomString } from 'helpers/generateRandomString';
```
wholesomedev: This is not 100% guaranteed to give you a unique ID every call. Wouldn't this create flakiness in the application at scale?

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -93,6 +95,17 @@ const Page = withRouter(
...
+                  logKey={notablePerson.slug}
+                  event={{
+                    type: 'PAGE_LOADED',
+                    payload: {
+                      url: history.createHref(location),
+                      timestamp: new Date(),
+                      isServer: __IS_SERVER__,
+                    },
+                  }}
+                />
```
wholesomedev: I think one issue with this approach is that child components get mounted before their parents. So this component could get mounted and it would dispatch the correct logs then one of its siblings could throw and the parent component would not get completely mounted. This could result in having inaccurate logs.

Also, is it possible to decouple logging from the markup of the components? I feel if the components weren't aware of logging, we would have a cleaner implementation? Not sure how to achieve that though...

src/app/helpers/compose.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @example
+ * ```typescript
+ * <button onClick={compose(onClick, logClickEvent)} />
+ * ```
+ */
+export function compose<T>(
+  ...fns: Array<undefined | ((...args: T[]) => void)>
+) {
+  return (...args: T[]) => fns.forEach(fn => fn && fn(...args));
+}
```
wholesomedev: This can be confusing since `compose` generally means something like 

```
foo(bar(2)) === compose(foo, bar)(2)
```

src/app/components/Loggable/Loggable.ts
```diff
@@ -0,0 +1,17 @@
...
+  children(childProps: ConnectableChildProps): JSX.Element | null;
+};
+
+const Loggable = ({ children, ...rest }: ConnectableProps) => {
+  return children(rest);
+};
+
+const ConnectedLoggable = connect(undefined, { log })(Loggable);
+
+export { ConnectedLoggable as Loggable };
```
wholesomedev: I'm not sure I understand this component correctly. Does it only provide a `log` function? If so, why not just import the `log` function directly from `store/features/logging/actions` in the component that needs that function?

src/app/components/Collapsable/Collapsable.tsx
forabi: We don't need this to be globally unique, just unique enough so that it does not conflict with other IDs on the same page.

src/app/components/Loggable/Loggable.ts
forabi: I'll probably go through various iterations before I settle on one 😄

src/app/helpers/compose.ts
forabi: Renamed to `callAll`

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: > I think one issue with this approach is that child components get mounted before their parents. So this component could get mounted and it would dispatch the correct logs then one of its siblings could throw and the parent component would not get completely mounted. This could result in having inaccurate logs.

I updated `LogOnMount` so that it wraps the page components. This guarantees that the call is made after the children's `render` methods have been called.

> Also, is it possible to decouple logging from the markup of the components? I feel if the components weren't aware of logging, we would have a cleaner implementation? Not sure how to achieve that though...

We could extract this to a generic `Page` component that logs when the page has loaded, but I think we'll need the components to be aware of logging sooner or later. What if we want to log page failures? Like in the `NotablePerson` page above:

```typescript
if (isErrorResult(result) || !result.value) {
  return (
    <LogOnMount
      logKey={this.props.slug}
      event={{
        type: 'PAGE_LOAD_FAILED',
        payload: {
          url: history.createHref(location),
          timestamp: new Date(),
          isServer: __IS_SERVER__,
        },
      }}
    >
      <MessageWithIcon
        title="Are you connected to the internet?"
        description="Please check your connection and try again"
        button={
          <LinkButton to={location} onClick={forceReload}>
            Reload
          </LinkButton>
        }
        icon={warningIconComponent}
      >
        <Status code={500} />
      </MessageWithIcon>
    </LogOnMount>
  );
}
```

src/app/components/Collapsable/Collapsable.tsx
wholesomedev: There is still a teeny tiny chance of collision right? 😄 

What if instead you have a namespaced counter? 

```js
const generateId = counter('collapsable');

generateId() // collapsable-1
```

forabi: This is what lodash `uniqueId` does:

```typescript
// User 1 requests this page
uniqueId('collapsable_'); // collapsable_1

// User 10000 requests this page
uniqueId('collapsable_'); // collapsable_10000
```
This number will keep growing indefinitely _on the server_ until the process restarts, which might be okay.

> There is still a teeny tiny chance of collision right?

The randomness is as good as V8's `Math.random` implementation. It should be fine I guess?

Anyway, your call!

src/app/helpers/callAll.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Creates a function that calls all provided functions
+ * with the same arguments passed to the composed function.
```
wholesomedev: Can we remove the word `composed` here?

src/app/typings/global.d.ts
```diff
@@ -1,6 +1,28 @@
...
+/**
+ * The base URL of the current app server,
+ * i.e. https://hollowvese.com for production or
+ * http://localhost:3001 for development
+ */
+declare const __BASE__: string;
+
+/** @example https://api.hollowverse.com/graphql */
+declare const __API_ENDPOINT__: string;
+
```
wholesomedev: 👍

src/app/components/Collapsable/Collapsable.tsx
wholesomedev: Following Discord discussion, I think this is fine.

forabi:
@wholesomedev I've experimented with "passive" logging, i.e. the components are completely unaware of logging. You can take a look at the code to see how I implemented that.

While I like the idea that all logging logic is contained in one place and that the components do not need to be modified to add logging, it looks like the logging epic will have to know which data every page needs. In addition to the maintenance burden, this approach is problematic because there is no deterministic way of knowing if the current page is a notable person's page before we have the notable person data, because the `/:slug` pattern could also match `/search`, `/privacy` and `/about`.

src/app/client.tsx
```diff
@@ -41,3 +42,15 @@ if (module.hot) {
...
+window.onerror = (message, source, line, column) => {
+  store.dispatch(
+    unhanldedErrorThrown({
+      message,
+      source,
+      line,
+      column,
+    }),
+  );
+};
```
wholesomedev, line 59: How are we going to get the location of the error in the source, noncompiled code?

src/app/hocs/WithData/WithData.tsx
```diff
@@ -12,6 +12,12 @@ type OwnProps<Key extends ResolvedDataKey = ResolvedDataKey> = {
    */
   dataKey: Key;
 
+  /**
+   * (Optional)
+   * The page path for which this data request is triggered, useful for logging
+   */
+  forPage?: string;
```
wholesomedev, line 19: This makes the implementation of `WithData` specific to Pages. Wouldn't it be better if `WithData` accepted arbitrary values to send along instead?

src/app/store/features/logging/epic.ts
```diff
@@ -0,0 +1,77 @@
...
+    .ofType('SET_STATUS_CODE')
+    .withLatestFrom(action$.ofType(LOCATION_CHANGE))
+    .map(([setStatusCodeAction, locationChangeAction]) => {
+      const url = createPath(
+        (locationChangeAction as Action<typeof LOCATION_CHANGE>).payload,
+      );
+      const statusCode = (setStatusCodeAction as Action<'SET_STATUS_CODE'>)
+        .payload;
+
+      if (statusCode === 200 || statusCode === 404) {
```
wholesomedev: What about `301`s or other statuses?

src/app/client.tsx
forabi: We already have source maps configured in production, browsers will just do the right thing here and report the error location after applying the source maps.

src/app/store/features/logging/epic.ts
forabi: This could be written `statusCode < 500`, but I was wondering what should we do with redirects? Should we ignore the request at this point and let it be handled after the redirection just like any other request?

src/app/hocs/WithData/WithData.tsx
forabi: If we allow arbitrary data, we will lose the benefits of static typing. That said, we could add more optional parameters, like, say, `forComponent`, and use type guards to figure out if the request is for a page or something else.

src/app/store/features/logging/epic.ts
wholesomedev: i think redirection is an interesting event that should be logged

src/app/client.tsx
wholesomedev: I thought source maps only get downloaded when the developer opens console.

If they will also get downloaded for onerror handler, then that’s good :)

forabi: > I thought source maps only get downloaded when the developer opens console.

I made a synthetic error to throw 5 seconds after page load, and logged the error event. I tested with DevTools open and closed. Looks like the error location is not mapped to the original source code regardless of whether DevTools is open or closed. So getting the error position is going to be more involved than I first thought. I found a few libraries, one of which is Mozilla's [`source-map`](https://github.com/mozilla/source-map/) which can get us the original error position, we can process this info on the server after we get the log event, transform it so that it points to the original source code, and send it to Sumo Logic.

forabi:
Here's what we are logging so far:

* Page load failed
* Page load succeeded
* Page redirected
* Unhandled JavaScript errors (although the error location is not source-mapped)
* ~Notable Person page visited through search (we log the search query and the NP's slug)~

src/app/store/features/logging/epic.ts
forabi: I've added a new action to log page redirections

wholesomedev:
@forabi I think we need to log search independently from NP page view. 

With regard to search, we can log the following events:

* User is viewing the search page (this can be a page view event, like the NP page view event)
* User has performed a search (we can define this event as being the event when user clicks on an item on the search page)

src/app/hocs/WithData/WithData.tsx
wholesomedev: If we're indeed going to be making community-reusable components, don't we need to make these components as generic as possible?

src/app/store/features/logging/epic.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+
+      return pageLoadFailed(url);
+    });
+
+  const loggableActions$ = action$.filter(shouldActionBeLogged);
+  let flushOnUnload$ = Observable.empty();
+
+  if (!__IS_SERVER__) {
+    flushOnUnload$ = Observable.fromEvent(window, 'unload');
```
wholesomedev: As you said, we need to think of our code as being universal to both browser and Node. By doing this, I feel we're using a browser concept that cannot be replicated on the server which could result in inconsistent and/or unreliable logging.

wholesomedev: Also, I don't think we should be waiting until the user is exiting the application to send out all the logs.

forabi: We are not waiting until the user closes the page. This is to handle logs that have been accumulated but not sent because they are less than 10. So the logs are sent when _either_: 

* 10 events are accumulated
* User navigates away

The only guaranteed way to send network requests on page unload is to use `navigator.sendBeacon`. Asynchronous requests using `fetch` or `XMLHttpRequest` that are sent on page `unload` are ignored by the browser.

wholesomedev: Oh I see. I misread the code. 

I think a better approach than waiting for 10 events could be to set a timer that dispatches events every 2 seconds or so

forabi: Using a timer is polling, I don't like polling. 😄

wholesomedev: What's wrong with using polling in this instance? :)

forabi: Why use polling when we can use get the observable to push the data instead? I don't see any benefits to polling over this here.

wholesomedev: Yeah, I guess you're right. If the `window.unload` handler works reliably, there isn't much benefit to using the timer over this

src/app/hocs/WithData/WithData.tsx
forabi: Yeah, you're right, but we can worry about that when we start moving the components to separate npm packages 😄

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -103,7 +74,9 @@ class Page extends React.PureComponent<Props> {
                       className={classes.placeholder}
                       icon={<SvgIcon {...searchIcon} />}
                       title="No results found"
-                    />
+                    >
+                      <Status key={searchQuery} code={404} />
```
wholesomedev: Algolia is continuously searching as the user is typing, does that mean this code could potentially get triggered while the user is still in the middle of typing?

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -115,14 +88,22 @@ class Page extends React.PureComponent<Props> {
                   value.hits.length === 1
                 ) {
                   return (
-                    <Status code={301} redirectTo={`${value.hits[0].slug}`} />
+                    <Status
+                      key={searchQuery}
+                      code={301}
+                      redirectTo={`${value.hits[0].slug}`}
+                    />
```
wholesomedev: I'm not sure this is a good idea. `301` means "moved permanently". The page hasn't really moved permanently. I think what would be better would be to show the search page with one result and then the user can click on that result to go to the NP's page.

src/app/store/features/logging/epic.ts
wholesomedev: It occurred to me that in the very unlikely scenario that we need to send 10s of events in quick succession, the buffer behavior may be less optimal than a 2 second timer that would make sure logging gets dispatched no more than once every 2 seconds. Anyway, for now, we don't have that use case. We can revisit if that changes.

With regard to the server-side, will we ever need to flush events on the server side?

src/app/store/features/logging/epic.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  // Send interesting actions to log endpoint
+  const logInterestingEvents$ = loggableActions$
+    .share()
+    // Send batch of logs when either 10 log events
+    // are accumulated, or the user navigates away
+    .bufferWhen(() => loggableActions$.bufferCount(2).merge(flushOnUnload$))
```
wholesomedev: Are we going to buffer 10 or 2? The comment might need to be updated.

src/appServer.ts
```diff
@@ -23,6 +24,8 @@ const appServer = express();
 
 appServer.use(...securityMiddleware);
 
+appServer.use('/log', logEndpoint);
```
wholesomedev: I wonder if it would make sense to move the logging functionality to an AWS Lambda.

wholesomedev: If you're interested in doing that, we can do it in parallel to this PR. If not, we can create an issue for looking into that.

src/app/pages/SearchResults/SearchResults.tsx
forabi: I think it makes sense to make an exception for the search page so that location changes to the URL query don't trigger additional page views.

src/app/store/features/logging/epic.ts
forabi: That's where `requestIdleCallback` should help. The browser will provide a deadline via the callback so that if the work can be chuncked and any remaining work can be processed on the next idle callback. However, it's up to the callback to respect this deadline. I'll need to look at the source code of `rxjs-requestidlecallback-scheduler` and see if the work is chuncked.

src/appServer.ts
forabi: One issue I can see with that is that the log endpoint will need to access the compiled Webpack bundles to read the source maps and look up the original error position. We should be able to access the source maps via HTTP, but we'll need to make sure to match the branch. I think I have a better idea that will help with this and has additional benefits: I have been considering moving all the compiled output of the client to S3 and serve it from a dedicated subdomain or use `files.hollowverse.com`, this way we can:

* Use Cloudfront as a CDN to serve our client bundle
* Let CF handle gzip compression for us, so we can drop the native `shrink-ray` dependencies. CF does not support brotli compression though.
* We can access the files from outside the deployed EC2 instance, so we can more easily move the logging logic to Lambda.

forabi: Actually, I'm not sure moving to S3 is worth the added complexity. There are files that do not have hashes in the filename, like `index.html` and this could lead to conflicts between deployed branches. We can figure out how to match the environment and access the source maps via HTTP.

src/app/pages/SearchResults/SearchResults.tsx
forabi: I've updated the code to only log the search page visit once. It's also more robust now as it uses `distinctUntilChanged` to ignore identical combinations of `LOCATION_CHANGE` and `SET_STATUS_CODE` so React re-renders won't trigger new page visit actions.

src/app/pages/SearchResults/SearchResults.tsx
forabi: Removed

src/app/store/features/logging/epic.ts
forabi: Updated the code to buffer 10 events

src/appServer.ts
forabi: I've implemented source-mapping of errors. The log middleware needed to be moved to the `./app` directory and bundled with Webpack so that we can use the action types and store helpers like `isActionOfType`. I wonder how easy it would be to use these types if we move the log endpoint to Lambda, and how the development workflow would work. Would we be able to test changes to the log endpoint while developing the app?

src/app/store/features/logging/epic.ts
forabi: It does not look like `rxjs-requestidlecallback-scheduler` respects the browser's specified deadline, although we are cooperating with the browser so that we don't block more important work like rendering and layout. I think this is acceptable for now.

forabi:
Error locations are now mapped to the source files. One remaining issue is that `PAGE_LOAD_SUCCEEDED` is not being dispatched on the server, which is weird because we're not doing anything special for the server regarding this action.

src/appServer.ts
wholesomedev: I don't really know. These are all things we would have to research and evaluate. I just think it would make sense to move logging to a separate micro-service (i.e. Lambda).

src/app/components/NavBar/NavBar.tsx
```diff
@@ -45,67 +43,61 @@ export const NavBar = class extends React.Component<
...
+                      <ConnectedSearchView />
+                    ) : (
+                      <div className={classes.logoViewInner}>
+                        <div className={classes.logoWrapper}>
+                          <NavBarLink
+                            title="Homepage"
+                            className={classes.logo}
+                            to="/"
+                          >
+                            <img src={textLogo} alt={title} />
```
wholesomedev, line 80: Deeply nested code tends to be harder to read because there's a lot of context to keep in mind. Are you okay with restricting our indentation levels to encourage simpler code? We could use these two

* https://eslint.org/docs/rules/max-depth
* https://github.com/yannickcr/eslint-plugin-react/pull/1260

forabi: That's always going to be an issue with JSX. TypeScript also make it less convenient to split the code into smaller functions because we would need to redefine the types again for each function. That said, I sure don't like deeply nested code. I've opened an issue for this.

forabi:
I've added a polyfill for `IntersectionObserver` to fix #242 and tested the change on iOS.

I've also added a workaround for browsers that do not support `sendBeacon` (including Safari), which is to use a _synchronous_ XHR. Safari does not trigger the `unload` event either, but triggers `pagehide` instead. The logs seem to be sent correctly on Safari iOS 11.

The synchronous XHR has one caveat though: it blocks the next navigation until the request is completed. So when users reload the page, move to a new page, or close the tab, they may perceive the page to be unresponsive.

forabi:
I deployed this branch to https://hollowverse.com/?branch=new-app, including the fixes for #248 #237 #243 #242.

forabi:
> One remaining issue is that `PAGE_LOAD_SUCCEEDED` is not being dispatched on the server

Unfortunately I couldn't find the root cause for this. I used remote Redux devtools to track the actions dispatched on the server and it looks like none of the actions that should be dispatched by the logging epic are being dispatched server-side. I thought this could be because we are buffering log requests and the store is destroyed for each page request before we get a chance to send the batched log events, but sending each log event immediately didn't work either. I suspect it's an issue in `react-redux-epic`. I've opened a new issue (#253) to track this.

`PAGE_LOAD_SUCCEEDED` is still being sent client-side as soon as the page loads though.

forabi:
I think this is ready for review. There's still a few things I need to take care of before we merge this, like switching the API instance to use the Aurora database instead of the development instance, and copying the database contents to Aurora, and while at it, I'd like to fix https://github.com/typeorm/typeorm/issues/1397 before we launch

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -72,7 +79,8 @@ const Page = withRouter(
               );
```
wholesomedev, line 79: Should we log the error case?

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -23,37 +23,44 @@ import { LinkButton } from 'components/Button/Button';
 import { RelatedPeople } from './RelatedPeople';
 import { withRouter, RouteComponentProps } from 'react-router';
 import { forceReload } from 'helpers/forceReload';
+import { DispatchOnLifecycleEvent } from 'components/DispatchOnLifecycleEvent/DispatchOnLifecycleEvent';
```
wholesomedev, line 26: Hmm, I thought Prettier wrapped things at 80?

src/app/store/createConfiguredStore.ts
```diff
@@ -27,18 +28,13 @@ const defaultInitialState: StoreState = {
       requestId: null,
     },
   },
+  alternativeSearchBoxText: null,
```
wholesomedev, line 31: What is the `alternativeSearchBoxText` for exactly?

wholesomedev (approved):
A few comments. Otherwise looks great. Look forward to launching and seeing the logs! 😄

src/app/store/createConfiguredStore.ts
forabi: It's used to display the notable person name in the navbar when the user scrolls down

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: It _is_ logged because we are setting the status code and combinations of status code and location change actions are mapped to page load events (failure/success)

forabi: You can check by blocking the requests to the API in Chrome DevTools and see `PAGE_LOAD_FAILED` dispatched in Redux DevTools.

screenshot from 2018-01-22 09-07-44

forabi:
I've tracked the issue regarding trimmed spaces at the end of editorial summary text nodes to https://github.com/typestack/class-sanitizer/issues/8, it might be a bug in `class-sanitizer` or I might be doing something wrong.

src/app/pages/NotablePerson/NotablePerson.tsx
wholesomedev: Ah, okay, I didn't notice that.

I'm beginning to feel that centralization of logging and detecting patterns in streams to dispatch certain logs may actually be more complex and would require more engineering than we should take on. So perhaps mixing logging code with the rest of application code might be the simpler, more straightforward approach.

What do you think?

(We can rewrite stuff later. It doesn't have to be part of this PR.)

forabi:
I found the root cause of typestack/class-sanitizer#8 and I've forked it to fix the issue as well as add tests, quality checks and CI. https://github.com/forabi/class-sanitizer/tree/fix-common-property-name-confusion. I just need to configure CI and propose the changes to the original repository.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: I think each approach has pros and cons:

* On the one hand, having this logic in the logging epic allows us to take advantage of RxJS operators and look at things happening in different parts of the application at once, but that also comes at the cost of increased complexity as you mentioned.
* On the other hand, moving the responsibility of dispatching the log events to the relevant components is simpler if we don't have many log events, but it also means that identical events may get dispatched multiple times for the same components if React/Preact re-renders that component for any reason, and in that case the component will need to be smart about what constitutes a unique log event, so we are back again to putting more logic in components and increasing the complexity of rendering.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: We enforce a max length of 120 characters (not 80) using TSLint and ESLint, not prettier.

wholesomedev: Yes, but Prettier formats code automatically if it exceeds 80 characters by default

forabi:
I've populated the production database with the data from the development instance but when trying to switch the API to use the production database, the server is failing to connect to Aurora, even though the load balancer's security group is allowed to communicate with the Aurora instance. I'm looking into it.

PR #231, opened 2018-01-06
Check for unsupported browser features, using Hollowverse browser usage data

forabi:


wholesomedev (approved):
Nice! 👍

package.json
```diff
@@ -81,7 +83,8 @@
     "last 2 Safari versions",
     "last 2 Edge versions",
     "last 2 Opera versions",
-    "not OperaMini all"
+    "not OperaMini all",
+    ">= 5% in my stats"
```
wholesomedev: Should you remove the other entries in this array now?

forabi: Removed

forabi:
@wholesomedev Looks like the dynamic "in my stats" browser query is causing some dependency to throw an error on build. I replaced this query with the resolved browsers list:

```
❯ yarn browserslist '>=5% in my stats'
and_chr 62
ios_saf 10.3
ios_saf 10.0-10.2
```

wholesomedev:
@forabi How often would we have to update this? Can't we replace it with a more dynamic value like "latest Chrome" and "latest iOS"?

forabi:
> @forabi How often would we have to update this?

The data I used for this is between `2017-01-01` and `2018-01-05`. We can update that every now and then, or write a script that does that for us.

> Can't we replace it with a more dynamic value like "latest Chrome" and "latest iOS"?

I don't understand. What would be the benefit? `browserslist` supports queries like "latest Chrome" and "last 2 Firefox versions", and that's what we were doing, but these are based on global browser usage data and don't reflect our usage statistics.

wholesomedev:
@forabi it looks to me like our usage statistics are the same as "latest 2 Chrome" and "latest 2 iOS Safari"?

forabi:
> @forabi it looks to me like our usage statistics are the same as "latest 2 Chrome" and "latest 2 iOS Safari"?

I see. So should we remove the custom statistics file altogether and use "last 2 Chrome versions" and "last 2 iOS Safari versions"

wholesomedev:
Yeah I think it’s better to have dynamic values instead of the custom stats which would require regular updating

Issue #253, opened 2018-01-21
Log events are not being sent server-side


forabi:

forabi:
Since we now have the website behind CloudFront, server-side logging and analytics won't reflect the actual page visits and metrics anyway.

PR #256, opened 2018-01-22
Improve cover image

### LINKED ISSUE #238: Improve cover image
We've decided to make the following modifications to the cover image:

* Limit its width to the width of the container
* Pick the dominant color, add gradient to it and use that as the cover image
* Add static gradient color as appears in `hollowverse/design`

forabi:
Closes #238


forabi:
I'm experimenting with extracting the vibrant colors of the image and using these in a gradient for the cover image background. Here's a comparison of old and new backgrounds for some pages. @wholesomedev I'd like some feedback :smile: 


screenshot from 2018-01-25 23-20-18
screenshot from 2018-01-25 23-20-58
screenshot from 2018-01-25 23-20-51
screenshot from 2018-01-25 23-20-37
screenshot from 2018-01-25 23-20-31
screenshot from 2018-01-25 23-20-24

wholesomedev:
Hmm, I think the new ones look good. But can we cap the saturation upper limit? Like what if we don't want background images to be too saturated? For example, in the case of Chris Martin, his older picture is a bit paler than his newer picture, and I'm thinking perhaps that is better? Anyway, the new one is generally better 👍. Can you also show what it would look like on the desktop?

By the way, do you think we might eventually want to cache these image processing computations some where rather than computing them for every request, or is it fine to have the client compute them since this happens in a service worker?

wholesomedev:
Actually, never mind what I said about saturation level stuff. I think the new images look better 😄

forabi:
> Can you also show what it would look like on the desktop?

This still needs some work, I don't like how the gradient is abruptly cut off at both sides, it would look much better if the gradient blended into the background at the edges.

localhost_3001_chris_martin 1

forabi:
> By the way, do you think we might eventually want to cache these image processing computations some where rather than computing them for every request. By the way, do you think we might eventually want to cache these image processing computations some where rather than computing them for every request, or is it fine to have the client compute them since this happens in a service worker?

The processing is performed client-side only for now, because when I used `<WithData />` to perform the computation server-side, the response would be sent before the computation is done, I think it's the same issue causing #253 and it probably has to do with `react-redux-epic`.

That said, I'd like to perform the computation server-side and memoize the results so that it's not performed for repeated requests, because when performed client-side, computation takes quite a long time, even on desktop. The good news is that it does not block the main thread because it's done in a worker.

By the way, it's not a service worker, just a standard web worker. Service Worker is a special kind of worker that can intercept network requests, and it's pretty new. Web Workers on the other hand can be used to perform generic computation operations in a separate thread and they have been available for years, with great [browser support](https://caniuse.com/#feat=webworkers).

forabi:
I'd suggest to use the extracted vibrant color for `<meta name="theme-color">`, this is used as the color of the tab in Chrome for Android, it looks super cool:

screenshot_20180124-171902

This requires a head manager library though (#261, which in turns depends on #263)

wholesomedev:
Looks nice! 👍

forabi:
Here is the latest iteration. I'm still not satisfied with the desktop version. I'm going to deploy this to `new-app` so you can check more pages

localhost_3001_katherine_webb iphone 6 1
localhost_3001_katherine_webb 1
localhost_3001_katherine_webb
localhost_3001_katherine_webb iphone 6

forabi:
and here's what it looks like on Chrome for Android

screenshot_20180205-204227
screenshot_20180205-204239

forabi:
The latest changes are up on https://hollowverse.com/Tom_Hanks?branch=new-app

I've still not found why yellow and blue take a lot of space on desktop.

wholesomedev:
I think it looks fine on desktop, and looks really good on mobile! 👍

wholesomedev (commented):
🎉

PR #259, opened 2018-01-26
Fix log events not being logged

forabi:
Looks like TypeScript didn't catch this.


forabi:
This proves that we need some kind of tests :smile:

wholesomedev (approved):

PR #261, opened 2018-01-26
Tell search engines not to index the new app for now, some SEO improvements

forabi:
* [x] Add head manager library
* [x] Set the title of the notable person page to that NP's name
* [x] ~Allow pages on the new app to be accessed directly using the new NPs' slugs~ (now in a separate PR, #262)
* [x] Add canonical URL


src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -100,6 +101,12 @@ const Page = withRouter(
 
             return (
               <div className={classes.root}>
+                <Helmet>
+                  <link rel="canonical" href={commentsUrl} />
+                  <title>{`${name}'${
+                    name.endsWith('s') ? '' : 's'
```
wholesomedev: I researched this recently and learned that for an individual, we always need to add `'s` whether the name ends with `s` or not.

http://www.thepunctuationguide.com/apostrophe.html

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -126,7 +133,7 @@ const Page = withRouter(
                 </article>
                 {notablePerson.relatedPeople.length ? (
                   <div className={classes.relatedPeople}>
-                    <h2>Other interseting profiles</h2>
+                    <h2>Other interesting profiles</h2>
```
wholesomedev, line 134: 😅

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -51,6 +52,9 @@ class Page extends React.PureComponent<Props> {
 
     return (
       <div className={classes.root}>
+        <Helmet>
+          <title>Hollowverse - Search</title>
```
wholesomedev: Should we use a consistent format of 

```
Foo - Hollowverse
```

?

wholesomedev (approved):
👍

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -100,6 +101,12 @@ const Page = withRouter(
 
             return (
               <div className={classes.root}>
+                <Helmet>
+                  <link rel="canonical" href={commentsUrl} />
```
wholesomedev, line 105: Perhaps this can be tackled as part of another PR, but I just wanna bring it up now:

We need to make an exception for the canonical URL for pages that we enable. For example, for Tom Hanks when we re-enable it, the canonical URL should be `/Tom_Hanks`, not `/tom-hanks`.

src/app/pages/SearchResults/SearchResults.tsx
forabi: Actually with React Helmet we can set a title template. I just updated the code.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: TIL! Updated

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: I see. We can check if the slug is in the redirection map and set the canonical URL accordingly.

PR #262, opened 2018-01-26
Allow pages on the new app to be accessed directly using new slugs

forabi:


wholesomedev:
I think we should hold off on this until we complete the `<head>` tag functionality.

forabi:
@wholesomedev This should be safe to merge now

wholesomedev (approved):

PR #270, opened 2018-01-29
Re-enable Tom Hanks page, allow new pages to be accessed directly

forabi:
* [x] Allow new pages to be accessed directly using new slugs
    * [x] Move `redirectionMap.ts`, `isNewSlug.ts`, redirection logic to the new app directory (`src/app`) so that we can use the redirection map to set the canonical URL and use the generated API types for `isNewSlug.ts`.
    * [x] Make sure the canonical URLs of pages that are not explicitly redirected is set to the old website's page URL
    * [x] Make sure the canonical URLs of pages that are explicitly redirected is set the new website's page URL
  * [x] Re-enable redirection for Tom Hanks's page
  * [x] Test that the redirection works
  * [x] Test that calling `next()` in server middleware falls back to the old website
  * [x] Allow whitelisted pages to be indexed
  * [x] Fix duplicated `<head>` tags

### DIFF
--- a/src/app/redirectionMap.ts
+++ b/src/app/redirectionMap.ts
@@ -1,0 +1,10 @@
+/**
+ * A map of old slugs to new slugs that are explicitly set to be redirected
+ * to the new website.
+ */
+export const redirectionMap = new Map<string, string>([
+  ['tom-hanks', 'Tom_Hanks'],
+]);
+
+/** A set of new slugs of pages that are explicitly redirected to the new website */
+export const whitelistedNewPaths = new Set(redirectionMap.values());
--- a/src/app/redirectionMap.ts
+++ b/src/app/redirectionMap.ts
@@ -6,5 +6,8 @@ export const redirectionMap = new Map<string, string>([
   ['tom-hanks', 'Tom_Hanks'],
 ]);
 
-/** A set of new slugs of pages that are explicitly redirected to the new website */
-export const whitelistedNewPaths = new Set(redirectionMap.values());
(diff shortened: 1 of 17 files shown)


forabi:
@wholesomedev Want to take a look? With #271 merged, I don't think there remains any issues blocking launch. This PR:

* Redirects Tom Hanks's page to the new app
* Sets the canonical URL of redirected pages to the new app URLs
* Allows new pages to be indexed _only if they are whitelisted for redirection_
* Allows accessing of new app pages directly via URL
* Prevents indexing of new app pages that are not explicitly whitelisted
* Uses the old website page URL as the canonical URLs for new pages that are not explicitly whitelisted

src/app/redirectionMap.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export const redirectionMap = new Map<string, string>([
+  ['/tom-hanks', '/Tom_Hanks'],
+]);
+
+export const whitelistedNewPaths = [
+  ...Array.from(redirectionMap.values()),
+  '/search',
+  // tslint:disable-next-line:no-unnecessary-callback-wrapper
+].map(path => pathToRegExp(path));
```
wholesomedev, line 15: Why can't you just do `map(pathToRegExp)`?

wholesomedev (approved):
LGTM! 👍

src/app/redirectionMap.ts
forabi: TypeScript says the signature/number of arguments does not match

PR #271, opened 2018-01-30
Force scrolling to top on navigation

### LINKED ISSUE #257: Force scroll to top on new page load
Due to the way React works, DOM nodes get recycled for the next navigation and the browser could keep the current scrolling position after the page content has changed. We should implement a generic `<ScrollToOnMount />` to fix this.

forabi:
Closes #257


wholesomedev (commented):
LGTM

PR #273, opened 2018-02-02
Lint GraphQL queries

forabi:
Using [`eslint-plugin-graphql`](https://github.com/apollographql/eslint-plugin-graphql) we can lint GraphQL queries in GraphQL files or in template literal strings in source code. This can be used to prevent usage of deprecated fields, ensure we are following best practices for caching... etc.

screenshot from 2018-02-02 13-19-36

* [x] Lint `*.graphql` files
* [x] ~Decide whether to allow or prevent usage of GraphQL queries in TypeScript files using template literal tags. If allowed, make sure to lint these queries as well.~ (looks like parsing inline queries requires `babel-eslint` parser, which we can't use because our code is in TypeScript)


wholesomedev (commented):
LGTM

PR #274, opened 2018-02-02
Add preconnect Resource Hint for the API endpoint

forabi:
From [Can I Use](https://caniuse.com/#feat=link-rel-preconnect):

> Gives a hint to the browser to begin the connection handshake (DNS, TCP, TLS) in the background to improve performance.

It's likely that the user would visit another NP page via "Other interesting profile" or search results, so this should help reduce the API query time.


wholesomedev (commented):
Nice!

PR #278, opened 2018-02-02
Prevent page jumping to top when JS is slow to load

forabi:
React Router triggers a location change event when JS loads, and `componentDidMount` on `<ScrollToOnMount />`  was being called after JS is executed on the client. If scripts are slow to load, the user could have scrolled down the page by the time JS is executed and the page would jump to top.


wholesomedev (commented):
Nice catch!

PR #279, opened 2018-02-02
Add runtime flag to disable server-side rendering

forabi:
* [x] Add `<link rel="preload" />` to generated `index.html`


src/app/helpers/memoizeOnSuccess.ts
```diff
@@ -1,20 +1,23 @@
...
 /**
  * Executes an asynchronous function and caches the result only if
  * the promise returned from the function does not reject.
  */
 export const memoizeOnSuccess = <T extends string, R>(
   fn: ((arg: T) => Promise<R>),
+  cacheOptions: Pick<Options, 'max' | 'maxAge'> = { max: 10000, maxAge: 43200000 },
 ) => {
-  const map = new Map<string, R>();
+  const cache = createLruCache<string, R>(cacheOptions);
```
wholesomedev, line 11: Just curious, why did you need to switch to LRU cache here? (I'm not really familiar with the LRU cache concept)

src/env.ts
```diff
@@ -1,29 +0,0 @@
...
-    return JSON.parse(fs.readFileSync(ENV_FILE_PATH, 'utf8'));
-  } catch (e) {
-    return {};
-  }
-}
-
-export const env = {
-  // Provide some default values in case the file does not exist
-  ...tryGetEnv(),
-};
```
wholesomedev, line 29: Why is this not needed anymore?

package.json
```diff
@@ -69,7 +69,7 @@
       "prettier --write",
       "git add"
     ],
-    "**/*.json": [
+    "**/!(package).json": [
```
wholesomedev, line 72: Finally! 😆

wholesomedev (commented):
LGTM!

package.json
forabi: It was so annoying 😂

src/app/helpers/memoizeOnSuccess.ts
forabi: I realized that too many requests are going to hit the cache, including requests for pages, static assets and images on the old website, so the cache could grow too large. Another benefit is that we can set a maximum age for the cache entries so if we add or remove a NP, we won't be holding to invalid cache entries indefinitely.

src/env.ts
forabi: The `X-Hollowverse-Actual-Environment` is now set in `server.ts` which is bundled by Webpack. In Webpack config, we define `__BRANCH__` and `__COMMIT_ID__` as global variables.

PR #283, opened 2018-02-04
Use relative paths for SVG icons

### LINKED ISSUE #235: Search icon doesn't show upon a hard reload on an NP page

forabi:
This is an attempt to fix #235


forabi:
Nope! Doesn't fix it!

PR #286, opened 2018-02-04
Re-enable svg-sprite-loader

forabi:
So I looked at the source code of `svg-sprite-loader` and I have a theory on why #235 happens and particularly why it happens inconsistently.

I've learned the hard way that whenever something happens sometimes but not everytime, it's almost always because of a race condition, and race conditions happen because of shared global state.

`svg-sprite-loader` has two components:

* The loader itself
* A helper plugin that hooks into Webpack events and writes the sprite file after collecting all the imported SVGs

We were using two Webpack configurations, one for the server and one for the client, but they were both based on a common configuration file (`webpack.config.common.js`). For most use cases, this is fine, but we were sharing the _same instance_ of  `SpriteLoaderPlugin` between the two configurations. Classes are stateful by design, so it's very likely that if we use a single instance of a plugin that needs to maintain some internal state, bad things are going to happen.

This PR changes the Webpack configuration files to instantiate an entirely new `common` Webpack config object for each configuration, so that we get an isolated plugin instance for each configuration.

This is still only an educated guess on what caused the issue, based on my analysis of the code in `svg-sprite-loader`. If the issue happens again, then I'm completely wrong :sob:

### DIFF
--- a/package.json
+++ b/package.json
@@ -258,6 +258,7 @@
     "stylelint": "^8.4.0",
     "stylelint-config-standard": "^17.0.0",
     "stylelint-no-unsupported-browser-features": "^1.0.0",
+    "svg-sprite-loader": "^3.6.2",
     "svgo": "^1.0.3",
     "svgo-loader": "^2.0.0",
     "ts-loader": "^2.3.4",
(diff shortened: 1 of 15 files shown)


forabi:
Since the issue was happening inconsistently, we'll need to merge this and keep an eye on the issue to see if it happens in future builds.

wholesomedev (commented):
LGTM

PR #291, opened 2018-02-06
Re-enable address bar bouncing behavior on Safari for iOS

forabi:
This is causing another issue with Safari, which causes the page to be stuck and prevents the user from scrolling down when the initial touch to start scrolling is located at the topmost part of the page until the touching finger is released ([as can be seen here](https://www.dropbox.com/s/hw8f7mjm6z3tu0g/File%2006-02-2018%2C%205%2033%2008%20PM.mov?dl=0))


wholesomedev (commented):
LGTM

Issue #281, opened 2018-02-03
Document architecturally-significant files and folders in our codebase


wholesomedev:
We need to start documenting architecturally-significant files and folders in our codebase. 

It's not clear whether we should use in-code documentation or a Wiki or perhaps a tool that generate external documentation from in-code documentation.

wholesomedev:
The thought behind this issue is to highlight the architecturally significant file to our SSR architecture. Since we removed SSR, this issue is no longer relevant. Closing...

Issue #244, opened 2018-01-07
Document our architecture


wholesomedev:
Once we settle on an architecture that we're happy with and that we think will serve us for a good while, we should document that with diagrams and such...

PR #293, opened 2018-02-07
Do not cache dynamic pages between deployments

### LINKED ISSUE #287: Reloading a page sometimes happens with missing CSS styles
Relevant discussion from Discord:

```
[12:39 PM] mk: @fwz There's a weird issue that I've occasionally noticed, which is that: other than the top navbar, the page seems to be missing its CSS styles and things look messed up.

It seems to be happening when I try to load the page for the first time in a couple of hours. But once the page is loaded for the first time, I can't reproduce the issue any more. I've experienced this on iOS Safari.

Any suggestions on how I might be able to reproduce this? Should I create an issue for it?(edited)
[12:41 PM] fwz: I never encountered this. Please feel free to open an issue and I'll try to reproduce it
[12:41 PM] fwz: I would probably start by clearing the entire cache of the website and loading the page
[12:42 PM] fwz: Do the styles get fixed eventually when the page finishes loading?
[12:42 PM] fwz: Or do you have to reload?
[12:42 PM] mk: @fwz No, they don't get fixed. I have to reload.
[12:43 PM] fwz: That's weird.
[12:44 PM] fwz: I think this might be because of cache mismatches? For example, /Tom_Hanks is cached for 12 hours but it references stylesheets that were replaced by another deployment?
[12:46 PM] fwz: and it could be happening in Safari because it does not support the immutable directive?
[12:46 PM] fwz: Not sure if that makes sense. Just saying
[12:47 PM] mk: @fwz any suggestion on how to confirm/disprove this theory?
[12:49 PM] fwz: Not sure where to start
[12:49 PM] fwz: There might be a simpler explanation though
[12:50 PM] fwz: Do you keep the tab open in Safari and go back to it later and you encounter the issue?
[12:50 PM] fwz: It could be that iOS is caching the page but only the markup, I sometimes see that behavior in Firefox
[12:51 PM] mk: @fwz It happens when I reload (the tab may or may not already be open, I don't remember. Most likely it's already open)
```

forabi:
This PR splits the `X-Hollowverse-Actual-Environment` header to two headers:

* `X-Hollowverse-Actual-Environment-Branch`
* `X-Hollowverse-Actual-Environment-Commit-ID`

The latter is used as a field in the `Vary` HTTP header, so this basically tells browsers not to use cached pages if the commit ID of the environment differs

However, the `Vary` header should only be set on "dynamic paths", like notable person pages. Varying all paths on the commit ID nullifies any value of using reproducible filenames between builds. If `icons.621a6cac01f0fe001b3523f241f44bc6.svg` does not change between builds, we should not prevent the browser from using the cached version.

The `X-Hollowverse-Actual-Environment` is also intentionally split into two headers because only the commit ID is important for `Vary`, if a user moves from `beta` to `master` but the commit ID does not change between the two environments, we should ot prevent the browser from using the cached version either.

Why not `Vary` on the `branch` cookie?

From [Best Practices for Using the Vary Header](https://www.fastly.com/blog/best-practices-using-vary-header/)

> However, sometimes cookies are used for A/B testing purposes, in which case it's a good idea to Vary on a custom header and leave the Cookie header intact. This avoids a lot of additional logic to make sure the Cookie header is left for URLs that need it (and are probably not cacheable).

This may or may not (fix) #287


wholesomedev (commented):
It's hard to keep in mind how our architecture is wired together. I'm embarrassed to admit, but I forgot how `X-Hollowverse-Actual-Environment` exactly works.

#244 and #281 will help a lot in that area.

LGTM!

PR #294, opened 2018-02-07
Do not remove URL parameters on search page

### LINKED ISSUE #292: Swiping back to search causes lag and inconsistent behavior on the Compass of Malicious Brokenness
As a more fitting (and easier to type) name for iOS Safari, I suggest we call it the Compass of Malicious Brokenness or COMB for short (I've added this to the glossary, tentatively. Happy to change it to something else. Other names that occurred to me: Compass of Evil, Broken Compass, Compass of Doom, Compass of Greed, etc.)

So here's the behavior I'm noticing on COMB:

1. Go to search page
1. Search for an NP
1. Tap on NP to go to NP page
1. Swipe back to search page

**What's expected:**

The search page with the search results displayed should load.

**What actually happens:**

The page with the search results loads, but COMB is frozen. Then COMB unfreezes and displays the search page but without any results and the search field is not focused.

Funny thing, the freezing doesn't happen when the back button on the bottom navbar is used instead of the swipe gesture. However, the search page is always empty when it's navigated back into.

(P.S. I realize this might not be a problem with COMB specifically, but iOS Safari is still a nasty thing that deserves an equally nasty nickname, regardless of this issue.)

forabi:
This should fix #292


wholesomedev:
Will this also fix the freezing behavior? (I noticed freezing happens also when I swipe back and forth between NP pages sometimes)

forabi:
> Will this also fix the freezing behavior?

Based on my testing, it looks like it will.

wholesomedev (commented):
👍

PR #295, opened 2018-02-09
Add integration tests

### LINKED ISSUE #254: Add integration tests
Should we add integration tests, as we had previously discussed on Discord?

forabi:
Closes #254


forabi:
I did a lot of experimentation with two different approaches to integrating Jest into our Webpack build process:

* Jest transforms source files directly using `ts-jest`: This is the simplest approach and lets us use all the features of Jest (source maps, mocks... etc.) and take maximum advantage of Jest parallelism. However, it requires us to create a mock file for each type of loader we are using (things like images, SVGs, URLs) and we also need to define all the globals (`__API_ENDPOINT__`, `__IS_SERVER__`) and polyfills that Webpack provides. This quickly gets tedious and requires a lot of code duplication and maintenance to keep Jest and Webpack configurations in sync.
* Webpack compiles the tests: using a Webpack configuration made specifically for tests, we build a version of the app (including the tests) that Jest can then use to run the compiled tests. Because Webpack does all the heavy lifting of transforming imports of images, SVGs and URLs, defining globals... etc., the jest configuration file is kept simple. All we have to do is point Jest to the compiled files and it runs the tests without having to do any transformations. Source maps still work correctly so Jest can easily collect code coverage, point to error locations in the source files... etc.

The trickiest thing with the second approach is getting Jest's watch mode to work with Webpack's watch mode -- Webpack will output a single file containing all the tests by default, and this file will get larger in size as we add more and more tests. This means that Jest will not be able to run the tests in parallel, and a change to a single source file could take a considerable amount of time to be reflected in Jest.

I tried creating a separate entry in Webpack config for each source test file, which worked and allowed Jest to run tests in parallel and minimize the overhead of compiling changed files, but that still does not detect _new_ test files added in the source directory because Webpack will only look at the entry configuration at startup. I found [`wildcards-entry-webpack-plugin`](https://github.com/8427003/wildcards-entry-webpack-plugin) which can inform Webpack of newly added entry files.

So here is what we get with the second approach:

* [x] Minimum Jest configuration, no need for transforms or mocks for non-JS imports
* [x] Watch mode: Change something in source files => Webpack re-builds => Jest re-runs the tests
* [x] Multiple tests files to be able to run tests in parallel and minimize re-building overhead
* [x] We make sure `node_modules` are not bundled so builds are very quick
* [x] Snapshots work
* [x] Test environment is as close as possible to the actual environment
* [x] Source maps work
* [x] Code coverage works


There are still some issues that need to be tackled though:

* [ ] When tests are run in watch mode, and a test file is deleted from the source directory, it does not get removed from the destination directory and this causes Jest to fail. We need to reflect the removal of entry files in the destination folder. I found [a Webpack plugin](webpack-clean-obsolete-chunks) that does something similar but it only works with chunks, not entries. I [opened an issue for this](https://github.com/GProst/webpack-clean-obsolete-chunks/issues/26).
* [ ] Mocking modules in Jest (i.e. `jest.mock('./path/to/file')`) would probably not work because Webpack rewrites paths to modules. [`jestpack`](https://github.com/60frames/jestpack) should help with this, although it looks like it's not actively maintained and I'm getting deprecation warnings in Webpack's output.
* [ ] Probably many things that I haven't thought about but we will probably hit in the future as we add more tests

wholesomedev:
@forabi I think this so far looks really good!

The webpack approach seems like a path with more resistance but might be a better long term approach. The ts-jest path on the other hand is more straightforward, but requires a lot of tedious setup.

The webpack approach maybe the better long term approach, but here's one concern: what if more features and deviations get introduced into Jest in the future that require us then to reconfigure our webpack (things like jest module mocking)?

The ts-jest approach on the other hand is more tedious, but it is easier to reason about.

Anyway, again, thanks for the awesome research and work that you've put into this so far. We can try either approach, and re-evaluate and restructure at any time.

forabi:
> what if more features and deviations get introduced into Jest in the future that require us then to reconfigure our webpack (things like jest module mocking)?

I'm not really sure. I guess that's something we can't avoid. Webpack is doing too many things for us so there will always be conflicts between Webpack and some other tool, and we'll have to find some way to make all these tools work together :smile:

forabi:
I spent the whole day yesterday trying to come up with a testing strategy that is both simple and covers a lot of the app functionality at once.

The official [`redux-observable` documentation](https://redux-observable.js.org/docs/recipes/WritingTests.html) and lots of discussions on their GitHub repo recommend testing individual epics instead of the entire store. That sounds like unit tests. I don't believe unit tests have much value when we have a lot of interactive pieces. A single epic may work exactly like it's meant to, but it may very easily break if another epic "swallows" or otherwise transforms actions that the first one expects.

Here's what I have in mind for the tests:

* Create app tree, provide a store instance
* Set the store to the desired state
* Render the page
* Make assertions about the page: page contains NP's name, editorial summary is rendered, ...etc.

I'm having a lot of difficulty with code splitting. `react-universal-component` is assuming we are running in a browser, which is true, but for some reason the wrapped pages are not being loaded and the components are stuck at "Loading...". I think I'm missing something, but I'd like to [force the universal components to resolve synchronously in Jest environment](https://github.com/faceyspacey/react-universal-component/issues/103) so it's one less thing to worry about. I've opened an issue on the package's repo to see if I can get help with that.

wholesomedev:
@forabi That sounds good. I agree with you that testing each epic in isolation would be inefficient, and would provide a poor guarantee of functionality.

The steps that you have in mind sound better to me, but one question:

How do you plan to "Set the store to the desired state"? (Ideally, that should happen through the dispatch of actions which trigger epics which further modify the store. That way, the tests would simulate the actual execution of the app as much as possible.)

forabi:
> How do you plan to "Set the store to the desired state"? (Ideally, that should happen through the dispatch of actions which trigger epics which further modify the store. That way, the tests would simulate the actual execution of the app as much as possible.)

 That's what I started with, but this turned out to be extremely difficult. We have lots of moving parts, actions triggered in response to other actions... etc. The biggest issue I faced with this is that there's no deterministic, clean way to know when the page has reached the "desired" state to run the assertions. I kept pursuing this and the best I could come up with is to define _in advance_ where we want to stop listening for actions. I did this by pushing a "listener" middleware at the end of the middleware chain and let the test `await` until this middleware `complete`s an outer observable:

```typescript
import { Subject } from 'rxjs/Subject';
const subject = new Subject():
const listenForSetResolvedData: Middleware = () => (next) => (action) => {
  if (isActionOfType('SET_RESOLVED_DATA') && isSuccessResult(action.payload.data)) {
    subject.complete();
  }
  return next(action);
};
const { store } = createConfiguredStore({ additionalMiddlware: [listenForSetResolvedData] });

mount(appTree);

store.dispatch(push('/Tom_Hanks'));

await subject.asObservable().toPromise();

expect(things);

```

What I didn't like about this is that we need to know which action to listen for in order to know that the page is in the desired state for tests. I read tons of discussions and that seems to be the most reliable approach.

I was actually writing a long description of what I was doing but then I realized that if we are going that route, we are really doing end-to-end tests. Pushing URLs is like simulating clicks and filling inputs with data.

Instead, why not take advantage of the fact that almost all of our state is contained in a single, centralized store? Why not just provide a snapshot of the app state in a particular point in time and see how the app would render? Here's what I settled for:

```typescript
import 'expect-more-jest';
import { createConfiguredStore, defaultInitialState } from 'store/createConfiguredStore';
import createMemoryHistory from 'history/createMemoryHistory';
import { getStatusCode } from 'store/features/status/reducer';
import { configure, mount } from 'enzyme';
import Adapter from 'enzyme-adapter-react-16';
import { createTestTree } from 'helpers/testHelpers';

configure({ adapter: new Adapter });

describe('Testing tests', () => {
  it('Notable Person page returns 200, displays NP details when found', async () => {
    expect.hasAssertions();

    const history = createMemoryHistory({ initialEntries: ['/Tom_Hanks'] });
    const notablePersonQueryResponse = {
      notablePerson: {
        commentsUrl: '',
        name: 'Tom Hanks',
        editorialSummary: null,
        mainPhoto: null,
        relatedPeople: [],
        slug: 'Tom_Hanks',
        summary: null,
      },
    };
  
    const { store } = createConfiguredStore({
      history,
      initialState: {
        ...defaultInitialState,
        resolvedData: {
          ...defaultInitialState.resolvedData,
          notablePersonQuery: {
            hasError: false,
            isInProgress: false,
            requestId: 'Tom_Hanks',
            value: notablePersonQueryResponse,
          },
        }
      },
    });

    const tree = createTestTree({ history, store });

    const wrapper = mount(tree);

    expect(wrapper.contains('Tom Hanks')).toBe(true);
    expect(getStatusCode(store.getState())).toBe(200);
  });
});

```

wholesomedev:
> why not take advantage of the fact that almost all of our state is contained in a single, centralized store?

So which parts of our codebase would not be covered if we do that, just the epics? What about the reducers, selectors, etc?

forabi:
> So which parts of our codebase would not be covered if we do that, just the epics? What about the reducers, selectors, etc?

The store that I used for the test above is exactly the same configured store that we use in the app. The only difference is the initial state. We make the minimum changes to the initial state and render the app. The app will still function as usual, and it may trigger other actions that go through the epics. The difference is that we don't wait for the app tree to get to the desired state, instead we "boot" the app in that state.

Reducers are part of the configured store so they would be tested as well. Selectors can be tested like in the example above (we test the store and the selector at once):

```typescript
expect(getStatusCode(store.getState())).toBe(200);
```

Anyway, I just realized that even with this approach, we might still want to wait for something to happen asynchronously, in that case we could go back to awaiting an action observable to complete.

Testing is painful :cry:

forabi:
Anyway, I agree that booting the app in the desired state means that the epics will be less covered.

wholesomedev:
What if we create a `waitToPass` utility function which would run an assertion and re-render the component repeatedly within 200ms timeout. If after 200ms, the assertion still hasn't passed, the test would be considered a failure?

I know it's a bit hacky, but I don't see why it wouldn't be stable...

Anyway, I just thought I'd throw this suggestion in our pool of potential solutions...

forabi:
> What if we create a waitToPass utility function which would run an assertion and re-render the component repeatedly within 200ms timeout. If after 200ms, the assertion still hasn't passed, the test would be considered a failure?

I've thought of this. In fact, there's already [a plugin for enzyme](https://github.com/etiennedi/enzyme-wait) to do this, and Jest will automatically fail the tests after a specified amount of time if no assertions are called.

It's definitely hacky but I think it might be good because we are testing the app at a high level and not having to depend on implementation details, like which action to wait for.

But that just feels like end to end tests. Navigating to pages, waiting for components to appear... etc.

wholesomedev:
In my opinion, it's okay for these tests to feel like E2E. For apps, E2E
tests are more effective than unit/integration tests. The downside of E2E
tests is that they are slow, unstable, and require an expensive environment
and setup to run. If we can have E2E-like tests without those downsides,
then that's a good thing.

forabi:
@wholesomedev Things are starting to look better :smile:, although there is some duplication and hacks. For example, I had to duplicate the app tree initialization code to replace dynamically-loaded components with their synchronously-resolved version. I also had to use `setTimeout(cb, 0)` and force update the tree to get the tests to pass.

```typescript
beforeEach((done) => {
      /* ... */
      wrapper = mount(tree);

      setTimeout(() => {
        wrapper.update();
        done();
      }, 0);
});
```

This is due to the way the event loop works in JavaScript. Using `setTimeout` will always schedule a "task" to be executed _after_ any _microtasks_. Promises are scheduled as microtasks, so `setTimeout` callbacks will always be delayed until all promises have settled. Jake Archibald has [a great article on this topic](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) and [he also gave a talk about the event loop at JSConf Asia](https://www.youtube.com/watch?v=cCOL7MC4Pl0) recently.

The tests have already discovered a bug in `<MessageWithIcon />` that was causing the `returns 404` test to fail. :sunglasses: 

screenshot from 2018-02-15 21-25-33

forabi:
I added tests for the search page

screenshot from 2018-02-16 00-01-22

wholesomedev:
This. Is. So. AWESOME!

I think you wired things together really nicely. I understand that there are hacks and that things are less than ideal, but I'm sure we can iron these out eventually.

I think `setTimeout` is fine too. I've always wondered if there were a way to wait for all immediately-resolved promises to finish. I'll checkout Jake Archibald's article.

I have a few minor comments, questions, and potential suggestions, but nothing substantial, so I'll wait for this to get merged first. Great stuff!

forabi:
I think this is in a pretty good shape now! Let's merge it! We can continue improving our test setup in the future :tada:

package.json
```diff
@@ -5,7 +5,14 @@
   "main": "index.js",
   "repository": "https://github.com/hollowverse/hollowverse.git",
   "scripts": {
-    "test": "npm-run-all -l --aggregate-output -s app/graphql/build -p validate-filenames lint",
+    "pretest": "npm-run-all -l --aggregate-output -s app/graphql/build -p validate-filenames lint",
+    "test": "run-p -l --aggregate-output app/test",
+    "app/test/webpack/dev": "NODE_ENV=test NO_TYPE_CHECK=1 yarn webpack --config src/webpack/webpack.config.test.js --watch",
+    "app/test/jest/dev": "jest --watchAll",
+    "app/test/dev": "npm-run-all -s clean/test -p 'app/test/*/dev'",
+    "app/test/webpack": "NODE_ENV=test yarn webpack --config src/webpack/webpack.config.test.js",
```
wholesomedev, line 13: What does `yarn webpack` do? How does `yarn` resolve `webpack`?

src/app/__tests__/notablePersonPage.ts
```diff
@@ -0,0 +1,140 @@
...
+  let wrapper: ReactWrapper<any>;
+  let store: Store<StoreState>;
+  let history: History | undefined;
+  let sendLogs: EpicDependencies['sendLogs'] | undefined;
+  let getNotablePersonResponse: EpicDependencies['getResponseForDataRequest'];
+  let initializeStoreAndTree: () => Promise<void>;
+
+  beforeAll(() => {
+    initializeStoreAndTree = async () => {
+      ({ store, history } = createConfiguredStoreForTests({
```
wholesomedev, line 26: What the hell? I never knew this was possible. Cool!

src/app/__tests__/notablePersonPage.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      // Force promises to settle by scheduling
+      // the following statements after `setTimeout`
+      await delay(0);
+      wrapper.update();
+    };
+  });
+
+  beforeEach(() => {
+    expect.hasAssertions();
```
wholesomedev, line 49: Can this be the default for every `it`?

src/app/__tests__/searchPage.ts
```diff
@@ -0,0 +1,192 @@
...
+      });
+
+      wrapper = mount(tree);
+
+      // Force promises to settle by scheduling
+      // the following statements after `setTimeout`
+      await delay(0);
+      wrapper.update();
+    };
+  });
```
wholesomedev, line 47: Can this function be abstracted somehow so that we don't have to put that comment about forcing promise resolution in every test suite?

src/app/__tests__/searchPage.ts
```diff
@@ -0,0 +1,192 @@
...
+          hitsPerPage: 10,
+          nbHits: 0,
+          nbPages: 1,
+          page: 0,
+          params: '',
+          processingTimeMS: 1,
+          query: 'Tom',
+        };
+
+        getSearchResultsResponse = createMockGetResponseForDataRequest(
```
wholesomedev, line 175: Can this be passed to `initializeStoreAndTree`? It's hard to read now because it looks as though you're initializing `getSearchResultsResponse` and nothing happens with it (you have to look at what `initializeStoreAndTree` is doing to realize that `getSearchResultsResponse` is being used by it.

src/app/store/createConfiguredStore.ts
```diff
@@ -1,15 +1,80 @@
...
+ * global polyfills.
+ *
+ * Instead of having epics import their dependencies and use them directly,
+ * `redux-observable` allows injecting dependencies into the epics. This helps
+ * with testing and portability, because we can now more easily replace the
+ * dependency implementation when creating the epics. This works well with
+ * TypeScript because it requires that implementations conform to the expected
+ * shape.
+ *
+ * @see https://redux-observable.js.org/docs/recipes/InjectingDependenciesIntoEpics.html
```
wholesomedev, line 45: 👍

src/webpack/shared.js
```diff
@@ -88,10 +96,14 @@ exports.createCssModulesLoaders = (isServer = false) => [
...
         '// This file is automatically generated.\n// Please do not modify it.\n',
 
-      // Shorten the class name in production bundles to save some bytes
-      localIdentName: isProd
-        ? '[hash:base64:5]'
-        : '[name]__[local]--[hash:base64:5]',
+      localIdentName:
+        // Do not scope the class names in tests, otherwise
+        // every change to CSS styles will require updating the
+        // snapshots.
```
wholesomedev, line 102: Just wanna point out that there are no snapshots now

wholesomedev (commented):
Fantastic!

package.json
forabi: Um, this is not actually needed. I was copy pasting from my terminal :sweat_smile: . `yarn webpack` means use the locally installed webpack bin (in `./node_modules/.bin`) instead of the globally installed one. Since we always run these commands using `yarn`, this command will be `yarn yarn webpack`, which is redundant.

src/app/__tests__/notablePersonPage.ts
forabi: I just discovered this yesterday. We need parenthesis around the statement, otherwise the JS parser thinks that `{` starts a new block.

src/app/__tests__/notablePersonPage.ts
forabi: I feel it should be, but in jest that does not seem to be the case.

src/app/__tests__/searchPage.ts
forabi: I would like to have some sort of context in the test runner where we can assign things to `this` and use them in `it` instead of polluting the outer scope. Jest by default does not, but I think we can use a different test runner.

src/webpack/shared.js
forabi: Yeah, I had some snapshots but ended up removing them. I don't like snapshotting the entire app tree, instead I'd like to keep snapshots of relevant elements only where it makes sense and rely more on regular assertions.

src/app/__tests__/searchPage.ts
forabi: This is now moved to a helper function `createTestContext` which is used in all test files.

src/app/__tests__/notablePersonPage.ts
forabi: It turns out we can move this to the test setup file. I just did :smile:

src/app/__tests__/searchPage.ts
forabi: I refactored a lot of code in #302, I think it's now much cleaner.

PR #296, opened 2018-02-12
Experiment with the new React context API

forabi:
The new context API moves context from an experimental feature of React to a first-class feature. I'm particularly interested in how the new API could improve our testing experience. Instead of having to tinker with Jest and Webpack and mock module files (which I really dislike), we could just swap the `value` prop of the `Provider` that wraps the `<App />` component with a mock implementation that we could create in the tests on the fly.

The new API works very well with static typing.

Note: this is using a polyfill for the new API, because the official API has been merged into React but not released yet.

### DIFF
--- a/src/app/client.tsx
+++ b/src/app/client.tsx
@@ -25,6 +25,7 @@ import { client as defaultApiContext } from 'api/client';
 const defaultAlgoliaContext = async () => import('vendor/algolia');
 
 export type AlgoliaContext = typeof defaultAlgoliaContext;
+export type ApiClientContext = typeof defaultApiContext;
 
 export const AlgoliaContext = createReactContext(defaultAlgoliaContext);
 export const ApiContext = createReactContext(defaultApiContext);
--- a/src/app/client.tsx
+++ b/src/app/client.tsx
@@ -4,7 +4,6 @@ import { ConnectedRouter as Router } from 'react-router-redux';
 import domready from 'domready';
 import { hydrate } from 'react-dom';
 import createBrowserHistory from 'history/createBrowserHistory';
-import createReactContext from 'create-react-context';
 
 import { App } from 'components/App/App';
 import { createConfiguredStore } from 'store/createConfiguredStore';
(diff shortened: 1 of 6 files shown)


forabi:
@wholesomedev I'd like to know what you think about this.

wholesomedev:
I think this by itself is fine. 

But I don't think this will scale well as a replacement for the ability to mock ES modules. It wouldn't make sense to create a context for everything we want to mock, right?

forabi:
> But I don't think this will scale well as a replacement for the ability to mock ES modules. It wouldn't make sense to create a context for everything we want to mock, right?

Well, for our own modules we could opt for a single context that has everything we need:

```typescript

const AppDependenciesContext = createReactContext({ apiClient, algoliaClient, /* ...otherDeps */ });

<AppDependenciesContext.Provider>
	<App />
</AppDependenciesContext.Provider>
```

and since everything is statically typed, we can still access what we need safely.

But I agree about scalability, we could still need to mock an entire module.

forabi:
I've cleaned up this PR from unrelated changes so it's easier to review.

src/app/client.tsx
```diff
@@ -11,6 +11,10 @@ import { StoreState } from 'store/types';
 import { unhanldedErrorThrown } from 'store/features/logging/actions';
 import { loadIntersectionObserverPolyfill } from 'helpers/loadPolyfill';
 import { HelmetProvider } from 'react-helmet-async';
+import {
+  AppDependenciesContext,
+  defaultAppDependenciesContext,
+} from 'appDependenciesContext';
```
wholesomedev, line 17: This API and naming is confusing. From `appDependenciesContext`, we're importing `AppDependenciesContext` and `defaultAppDependenciesContext`. Then we're initializing `AppDepdenciesContext.Provider` and passing `defaultAppDependenciesContext` as `value` to it.

Also, the way we create `AppDependenciesContext` is by calling `createReactContext` with `defaultAppDependneciesContext`.

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -51,116 +53,139 @@ const Page = withRouter(
...
-              return (
-                <MessageWithIcon
-                  title="Are you connected to the internet?"
-                  description="Please check your connection and try again"
-                  button={
-                    <LinkButton to={location} onClick={forceReload}>
-                      Reload
-                    </LinkButton>
+        <AppDependenciesContext.Consumer>
+          {dependencies => {
```
wholesomedev, line 57: If we're using React context, isn't there a better way to grab these dependencies other than render props? Perhaps at the construction of the component instead? I imagine a new person looking at the code and going "wtf?"

wholesomedev (commented):
I just think using render props for the dependencies context looks a bit weird and requires that you wait until your component renders in order to be able to create functions, which doesn't seem ideal. It would be better if we can grab context dependencies earlier in the component lifecycle.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: This is the exact same context API that will become part of React 16.3, it's also the first official React API to use a render prop. I think people learning React will have to learn the render prop pattern.

src/app/client.tsx
forabi: Yeah, it is confusing. The requirement for passing a default value for `createReactContext` is enforced for "type safety", although I'm not completely sure what that is supposed to mean. The `value` prop of the `Provider` is where we can override the default context value, which we don't need to for the client-side app tree, but we do for the server-side version. Passing a `value` is not optional though, even if it's the same value passed to `createReactContext`.

forabi: I think we can rename `defaultAppDependenciesContext` to just `defaultAppDepenencies`. It will be less confusing. What do you think?

wholesomedev:
Yeah, I agree. That would be an improvement.

src/app/client.tsx
wholesomedev: @forabi Yeah, I think that would be an improvement

PR #297, opened 2018-02-13
Restore git hooks, upgrade dependencies, clean up test files

forabi:
Git is getting stuck on pre-commit and pre-push hooks. I think the command that's responsible for this is `lint-js`.


forabi:
I think the reason for git getting stuck was that I was not ignoring the `__tests__` folder in `.eslintignore` and ESLint was taking a lot of time. It was printing thousands of issues in the compiled files.

.eslintrc.json
```diff
@@ -1,3 +1,31 @@
...
+        "env": "literal"
+      }
+    ],
+    "graphql/named-operations": [
+      "error",
+      {
+        "env": "literal"
+      }
+    ]
+  }
```
wholesomedev, line 30: If we run into a situation where we need these same rules in another project, maybe we could convert `eslint` files to Common JS modules so that we're better able to move things around. For now this is good.

wholesomedev (commented):
LGTM

PR #302, opened 2018-02-17
Refactor tests, fix CSS Module class names configuration

forabi:
Instead of having multiple variables in the outer scope of each test, this PR leaves only one variable, `context`, and adds a helper function, `createTestContext`, that simplifies setting up the test tree and store.

### DIFF
--- a/src/app/__tests__/searchPage.ts
+++ b/src/app/__tests__/searchPage.ts
@@ -1,103 +1,70 @@
-import { EpicDependencies } from 'store/createConfiguredStore';
-import createMemoryHistory from 'history/createMemoryHistory';
 import { getStatusCode } from 'store/features/status/reducer';
-import { mount, ReactWrapper } from 'enzyme';
 import {
-  createTestTree,
   createMockGetResponseForDataRequest,
-  createConfiguredStoreForTests,
+  TestContext,
+  createTestContext,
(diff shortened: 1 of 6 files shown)


src/app/__tests__/searchPage.ts
```diff
@@ -1,154 +1,128 @@
...
-          await delay(5000);
-
-          return searchResults;
-        }
+      context = await createTestContext({
+        createHistoryOptions: { initialEntries: ['/search'] },
+        epicDependenciesOverrides: {
+          getResponseForDataRequest: async payload => {
+            if (payload.key === 'searchResults') {
+              await delay(5000);
```
wholesomedev, line 80: I forgot to ask in the previous PR, but I can't quite tell what this is for? Will it delay tests by 5 seconds? Wouldn't that be bad?

wholesomedev (commented):
Yep, much cleaner. LGTM!

src/app/__tests__/searchPage.ts
forabi: No, we are not waiting. We are just making sure that the test gets enough time to stay in the "loading" state to make the assertions. If we were waiting for 5 seconds, Jest would have failed the test because it took too long.

PR #306, opened 2018-02-19
Add test coverage

### LINKED ISSUE #305: Add code coverage report
We need to add the ability to report our integration tests code coverage. Might also wanna publish that to https://coveralls.io/

forabi:
Closes #305


forabi:
I've set up coveralls, but two issues remain:

* Lines are not correctly re-mapped
* Only files imported in the test files are considered. This covers most of the codebase, but not everything. Server files, like `serverEntry.ts` and `createServerRenderMiddleware.tsx` are not covered, because these are not imported in the tests. Mocked functions are also not covered, because we test the mock implementation instead of the actual functions.

forabi:
The current test configuration cannot be used to test the server rendering functionality. We will need an additional Webpack configuration file to generate a server bundle for tests.

wholesomedev:
I think those issues are fine. We can create a separate GitHub issue to track these improvements.

forabi:
Cool! I merged this and opened #309 and #310 to track the remaining issues.

PR #308, opened 2018-02-20
Use the provider pattern to get the chunks names in server rendering middleware

forabi:
For consistency and compatiblity with React async rendering in the future, this PR removes the dependency on global state for `react-universal-component` and instead provides the chunk names using the provider pattern.

(Blocked on https://github.com/faceyspacey/react-universal-component/pull/106)

* [x] Upgrade `react-universal-component`
* [x] Remove `@ts-ignore`


wholesomedev (commented):
LGTM

Issue #309, opened 2018-02-20
Lines are not correctly re-mapped in code coverage reports


forabi:

forabi:
Well, comparing the _local_ reports of both approaches there does not seem to be any significant differences. However, it seems that codecov is messing up lines in some TSX files. I don't think there is something we can do now, but I think the code coverage now is more accurate than what we had when we were compiling the tests using Webpack. I'm going to close #440.

PR #311, opened 2018-02-20
Configure Babel to retain line positions in compiled output

forabi:
[The `retainLines` option](https://babeljs.io/docs/usage/api/) instructs Babel to maintain the source line position in the compiled output.

> Retain line numbers. This will lead to wacky code but is handy for scenarios where you can’t use source maps. (NOTE: This will not retain the columns)

This should help with #309, although ideally correctly configured source maps should not need this.


wholesomedev (commented):
LGTM

PR #314, opened 2018-02-21
Simplify server rendering middleware code

forabi:
Convert observable to promise and use await to make code more readable


wholesomedev (commented):
LGTM

PR #315, opened 2018-02-21
Remove code coverage badge for now

forabi:
From Discord discussion:

> mk: the badge being red on the README is not good. We don't wanna be desensitized to red alerts and such...so I think if we can't make it green, we should remove it.


wholesomedev (commented):
👍

PR #316, opened 2018-02-21
Refactor imports to use new TypeScript `esModuleInterop` option

forabi:
See https://github.com/hollowverse/common/pull/37


wholesomedev (commented):
This is a nice change in TypeScript. It'll help with TypeScript adoption for newcomers.

👍

PR #317, opened 2018-02-23
Break notable person page into multiple files

forabi:


src/app/pages/NotablePerson/NonErrorNotablePerson.tsx
```diff
@@ -0,0 +1,95 @@
...
+          inView ? (
+            <Card className={cc([classes.card, classes.comments])}>
+              <FbComments url={commentsUrl} />
+            </Card>
+          ) : null
+        }
+      </OptionalIntersectionObserver>
+    </div>
+  );
+};
```
wholesomedev, line 95: I think this is fine for this PR and we can keep it for now, but I don't think this `NonErrorPage` is a good pattern that we should replicate in other places.

Is there no convenient way in TypeScript to keep things in one file but have multiple render methods?

wholesomedev (commented):
LGTM!

src/app/pages/NotablePerson/NonErrorNotablePerson.tsx
forabi: I didn't feel too comfortable moving this to a separate file. Maybe we can move it back to a method on `NotablePerson`. TypeScript does not make this any less convenient than regular JS.

PR #318, opened 2018-02-23
Trim generated CSS string for NP cover gradient

forabi:
Convert this:
screenshot from 2018-02-23 13-14-49

To this:
screenshot from 2018-02-23 13-13-26


wholesomedev (commented):
Cool 👍

PR #321, opened 2018-02-25
Recombine NotablePerson component

wholesomedev:
I've recombined the Notable Person component.

I found two ways to split up the render method.

#### Approach no. 1

Use curried methods

```jsx
getErrorMessage = () => ({someProp}) => (
  <ErrorMessage someProps={someProps}/>
)
```

and then use that in the `render` method as such

```jsx
render() {
  const ErrorMessage = this.getErrorMessage();

  return (
    <WithRenderProps>
      {someProp => {
        return <ErrorMessage someProp={someProp} />;
      }}
    </WithRenderProps>
  );
}
```

#### Approach no. 2

The other way is to *not* use a curried method, but return the initialized element

```jsx
renderErrorMessage = (someProp) => (
  <ErrorMessage someProps={someProps}/>
)
```

and then use that in the `render` method as such

```jsx
render() {
  return (
    <WithRenderProps>
      {someProp => {
        return this.renderErrorMessage(someProp);
      }}
    </WithRenderProps>
  );
}
```

### Pros / cons

Approach no. 1 pros: feels more true to JSX, looks good. Cons: it's verbose, currying can be confusing, less performant(?)

Approach no 2 pros: less verbose, more performant(?). Cons: returning the result of a method in the `render` function such as `return this.renderErrorMessage(someProp);` can look awkward to new eyes.

I prefer approach no. 2. But this PR contains approach no. 1. I'm open to either approach.

Let me know what you think.


forabi:
What about a third approach where we simply access `this.props` in the class method? Why do we have to pass the props to the error message? It's not like we are going to re-use it in another page, and if we were, it'd make more sense to keep it in a separate file.

Anyway, I don't really have a preference. I'm fine with any of these as long as they make the code more readable and easier to reason about.

wholesomedev:
@forabi It's not possible to use `this.props` because these components need the render props, which are not available until the component itself renders.

I'll update this PR to use approach no. 2 👍

wholesomedev:
@forabi Oh, I see that in the case of `ErrorMessage`, the data is available in `this.props`. So in that case, yeah, I'll use that 👍

wholesomedev:
@forabi This is ready now.

forabi:
Looks good to me!

PR #325, opened 2018-02-26
Use single-direction margins + attempt to standardize some CSS values

### LINKED ISSUE #239: Use single-direction CSS margins
https://csswizardry.com/2012/06/single-direction-margin-declarations/

### LINKED ISSUE #240: Build a CSS framework
I'd like us to eventually extract our styles into a reusable CSS framework of some sort. We're already doing that to a degree, but I think there's room for improvement.

We can standardize our component spacing like margins, paddings, etc. We can standardize our font sizes, so that we only have a handful of different sizes that we reuse everywhere.

wholesomedev:
### Use single-direction margins

This PR implements single-direction margins. I think this helped a lot in cleaning up some tricky margin spaces. 

Now we're only allowed to use `margin-top` and `margin-left` in the codebase. `stylelint` will complain if someone tries to use `margin`, `margin-bottom`, or `margin-right`. Obviously, one can disable stylelint where needed.

### Attempt to standardize some CSS values

This PR also attempts to standardize some CSS values that we use in our UI, namely: spacing values for margins and padding, which I put in `_variables.scss` and font definitions, which I put in `_mixins.scss`.

I used these variables and mixins where ever possible.

## To-do

- [x] Move `stylelint` config to `common`
- [x] Update Jest snapshots

---

Closes #239, closes #240


src/app/components/Label/Label.module.scss
```diff
@@ -13,11 +11,9 @@
   border: 1px solid $color-border-light;
 
   &.small {
-    margin: 0 2px 2px 0;
-    padding: 0 8px;
+    padding: 0 $spacing-8px;
     height: 18px;
     border-radius: 9px;
-    font-size: 11px;
```
wholesomedev: I didn't bother to not break this file since we're not really using it for now. And most likely when we use it, a lot of it will probably be rewritten. Might be worth considering deleting this file altogether.

(Same goes for the `Post` related files)

src/app/components/MessageWithIcon/MessageWithIcon.module.scss
```diff
@@ -1,7 +1,7 @@
 @import 'variables.scss';
+@import 'mixins.scss';
 
 .root {
-  margin: auto;
```
wholesomedev, line 4: I think it would be better to use flex layout to center things rather than `margin: auto`, just for consistency and simplicity. 

Anyway, I don't think deleting this broke anything.

src/app/pages/NotablePerson/NotablePersonSkeleton.module.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable property-blacklist */
 @import 'variables.scss';
```
wholesomedev, line 2: I didn't try to change too much in this file.

I'll take some time later to see if I can build the same loading effect without the need for skeleton components.

src/app/_mixins.scss
```diff
@@ -19,3 +19,27 @@
...
+}
+@mixin font-20px {
+  font-size: 20px;
+  font-weight: bold;
+}
+@mixin font-26px {
+  font-size: 26px;
+  font-weight: normal;
+  line-height: 1;
+}
```
forabi, line 45: Wouldn't it be better if we used names like `font-large`, `font-medium`... etc. instead of pixels? Same for spacing variables.

src/app/components/Label/Label.module.scss
forabi: Agreed. I think we should remove unused files.

forabi:
Margins/paddings around the summary text are missing

screenshot from 2018-02-26 14-44-45

.stylelintrc
```diff
@@ -3,6 +3,7 @@
   "rules": {
     "plugin/no-unsupported-browser-features": [true, {
       "ignore": ["css-featurequeries", "outline", "css-appearance"]
-    }]
+    }],
+    "property-blacklist": ["margin-bottom", "margin-right", "margin"]
```
forabi: :+1: 

I think there are a lot of rules we can add to stylelint to improve our CSS. I didn't have much time to look into this recently, but I'll definitely come back to it.

forabi:
I updated the snapshots and fixed the missing paddings

src/app/_mixins.scss
wholesomedev: @forabi Yeah, I thought about this. 

I don't like `font-13px`, `font-16px`, etc, because they can be misleading. One might think that all `font-13px` does is that it brings `font-size: 13px;` when in reality it brings a few other things too.

On the other hand, a naming scheme like `font-large`, `font-medium`, etc, creates a hard-to-adjust scale. Like, what if after we create this scale, we decide to add a font between `font-medium` and `font-large`? 

I've seen projects where they do `font-paragraph`, `font-title`, `font-caption`, but these values end up becoming pretty arbitrary and it becomes hard to name new fonts.

What do you think of `font-small-11px`, `font-small-13px`, `font-medium-16px`, at least this one sort of indicates that there's more to the mixin than just pixel sizing?

For spacing, I'm not sure. We have 10 values so far. If we can bring them down to 7, we can use `xxs`, `xs`, `s`, `m`, `l`, `xl`, `xxl`, but then we wouldn't be able to insert more values.

I think the scheme `$spacing-10px`, `$spacing-15px`, etc is acceptable. I know the names seem very primitive but we still get some standardization/consistency because we have one file where we define our app spacing.

wholesomedev:
@forabi Thanks for fixing the padding (and updating the snapshots!). I forgot to test in a narrow viewport! 😅

src/app/_mixins.scss
forabi: > we wouldn't be able to insert more values.

I think this is a valid point. Each naming convention will have some kind of limitation. It's just that I didn't see mixins/variables named like this before. It's interesting. 

> What do you think of `font-small-11px`, `font-small-13px`, `font-medium-16px`, at least this one sort of indicates that there's more to the mixin than just pixel sizing?

I don't think it indicates clearly that we are setting other CSS properties. Might be better to stick to `font-11px` for simplicity.

I was planning to move the stylesheets to `rem`/`em` units, that's why I didn't like having the font size and unit in the mixin name.

Anyway, like you said, the goal is some level of consistency, so I think this is fine for now.

Probably off-topic, but I wish CSS/Sass in VS Code had the same level of code refactoring/navigation support it has for TS and JS.

wholesomedev: @forabi Cool. I'll keep `font-11px` for now. In the future, if we switch to other units, we can rename these.

> I wish CSS/Sass in VS Code had the same level of code refactoring/navigation support it has for TS and JS.

Yeah, I feel it should be possible, right? I wanna hover over a mixin and see its definition! I guess that's one area where CSS-in-JS is better than Sass 😜

wholesomedev:
This is ready to merge. No rush though.

forabi:
I removed the unused `Label` and `Post` components.

Issue #323, opened 2018-02-26
"Loading..." text flashes on non NP pages


wholesomedev:
How to reproduce:

1. Go to https://hollowverse.com/Tom_Hanks
1. Position your cursor on "Hollowverse" logo in the navbar
1. While keeping your eye on the top-left corner of the viewport, click on "Hollowverse" logo in the navbar

You will see "Loading..." text flashing in the top left corner of the view port under the navbar.

forabi:
We need to use either a generic loading component or a skeleton for the home and text pages

PR #331, opened 2018-03-04
Implement skeleton as part of original markup

wholesomedev:
This PR implements the NotablePerson component skeleton as part of the original markup.

As I was nearing the completion of this work, I felt that it wasn't as easy to implement as I had hoped. So with regard to ease of implementation, this approach is probably just as hard if not harder than the earlier approach.

This approach has another issue that may be considered an advantage or disadvantage: it is intrusive. I've had to pass the `isLoading` flag around and use it as a condition to determine the parts of the JSX to execute. This is good because it ties the skeleton rendering to the original markup, which provides better confidence that the two are in sync. But it's also bad because now we're committed to using skeleton as our loading indicator. And it was cleaner when the components didn't care as much about the loading state.

One benefit I suspect this approach gives us is possibly teeny-tiny bit faster page rendering? Since the original markup is used so that there are fewer repaints? I hope so! 😄

Anyway, happy to discuss, make further edits, or completely revert changes. I feel there is a lot of room for improvement still.

Oh, in this PR I also broke up the rendering functions of `NotablePerson` and `PersonDetails` further. I felt that really improved readability of those components. Let me know what you think.

#### To-do:

- [x] Not working on COMB. Needs fixing.


src/app/pages/NotablePerson/NotablePersonSkeletonData.ts
```diff
@@ -0,0 +1,8 @@
+export const NotablePersonSkeletonData = {
+  name: 'Lorem Temporibus',
+  summary:
+    'Lorem ipsum dolor sit amet, et sit nonumy appareat temporibus, legere scripserit et ' +
+    'has. Ea accumsan luptatum maiestatis notcum. ' +
+    '\nEoctus vulputate eum cu. Dicam insolens vix ea, populo explicari ut duo. Quo discere ' +
+    'delectus eu. Exerci prompta iracundia sit te.',
+};
```
wholesomedev: I don't like the existence of this file or its location, but I'm not sure what to do about it...

wholesomedev:
Now that I removed `NotablePersonSkeleton` component from `LoadableNotablePerson`, I suspect that another issue with this implementation is that if the user first lands on, say, the Privacy Policy page, then navigates to a Notable Person page, I think they will see the "Loading..." flash (#323) momentarily before they see the skeleton? 

We would need to create a generic loading component to use in place of all loadable components I would think...

forabi:
I just want to point out that the `NotablePersonSkeleton` component serves two purposes:

1. Show a loading indicator while the actual `NotablePerson.tsx` is loading. The actual page is pretty heavy as it contains most of the third-party libraries we use.
2. Show a loading indicator while data for a notable person is loading.

We need to make sure 1. is still achieved. This means we cannot have the skeleton in the same file as the actual component.

wholesomedev:
@forabi yeah, I only realized that after I finished implementing the new skeleton.

I can’t think of a way to have the skeleton implementation be part of the original markup and at the same time serve as a loading indicator for the `LoadableNotablePerson` component.

I’m thinking we should create a generic loading component for all Loadable pages. If we then transition between the different loading states with nice fading effect, it should look fine.

What do you think?

forabi:
> I’m thinking we should create a generic loading component for all Loadable pages. If we then transition between the different loading states with nice fading effect, it should look fine.

I kind of liked the skeleton. Using the same skeleton for loading both the component and the data feels seamless from the user's point of view. I imagine if we use a generic spinner for loading the component and then use the skeleton for data it would be confusing. For a moment, the user would think that loading has finished, but they would see yet another loading state.

We should use one of the two loading indicators exclusively.

wholesomedev:
@forabi Hmmm, okay, let me try to refactor things and see what I can do. If all fails, I'll reinstate the old approach.

wholesomedev:
@forabi I've extracted out a `NotablePersonBodySkeleton` (awful name, need to come up with something better).

The purpose of this component is to be simple, stateless, and make it possible to render both a full body and a skeleton body.

What do you think?

src/app/pages/NotablePerson/NotablePersonBodySkeleton.tsx
```diff
@@ -0,0 +1,35 @@
...
+type Props = {
+  notablePerson?: NotablePersonQuery['notablePerson'];
+  editorialSummary?: JSX.Element;
+};
+
+export const NotablePersonBodySkeleton = ({
+  notablePerson,
+  editorialSummary,
+}: Props) => {
+  const dataSource = !notablePerson
```
forabi: If `notablePerson` is `null` it means the notable person is not found in the database. We should prefer strict comparison (`notablePerson === undefined`) to avoid any ambiguity.

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -101,37 +126,49 @@ const Page = withRouter(
...
-              ) : null
-            }
-          </OptionalIntersectionObserver>
-        </div>
+        </>
+      );
+    };
+
+    renderBody = (notablePerson?: NotablePersonType) => {
+      if (!notablePerson) {
```
forabi: Same here.

src/app/pages/PrivacyPolicy/LoadablePrivacyPolicy.tsx
```diff
@@ -1,5 +1,7 @@
 import universal from 'react-universal-component';
+import { NotablePersonBodySkeleton } from 'pages/NotablePerson/NotablePersonBodySkeleton';
 
 export const LoadablePrivacyPolicy = universal(import('./PrivacyPolicy'), {
   key: module => module.PrivacyPolicy,
+  loading: NotablePersonBodySkeleton,
```
forabi: You added this to the wrong file.

forabi (commented):
I think this approach is good because it streamlines and deduplicates our components and it can be applied to any page. It's very similar to what [`react-loading-skeleton`](https://github.com/dvtng/react-loading-skeleton) does.

One minor issue is that because image URLs are broken in the skeleton state, browsers will show the "broken image" placeholder for a brief moment while the data is loading. Not sure what the best way to work around this is (tweaking visibility? Not going to work)

src/app/pages/PrivacyPolicy/LoadablePrivacyPolicy.tsx
forabi: I just fixed it.

forabi:
I forgot to mention that we should probably add `aria-hidden={isLoading}` to the root of the skeleton to avoid screen readers from reading the invisible placeholder text.

wholesomedev:
@forabi I've addressed the PR comments and made some CSS adjustments. I think this is ready to merge now.

Also, tested on Safari. With the new implementation, it seems to work fine.

wholesomedev:
@forabi Actually, I haven't addressed the broken URL issue. Let me see what I can do about that

wholesomedev:
@forabi Okay, so the way I've dealt with the missing image URL is by creating an empty svg (by manually editing the svg code), and used that as the image for the skeleton state.

forabi:
@wholesomedev Cool! That's similar to what I had in mind (although I was thinking of using a transparent 1px PNG, does not matter I guess). I'll review this more thoroughly tonight and see if something needs to be tweaked.

forabi:
Looking good! I'll go ahead and start a similar PR for the search page. We can later look into whether we want a generic loading component or similar skeletons for the other pages.

wholesomedev:
Awesome!

This skeleton component thing is easy, so feel free to leave it up to me. I
can take it up on weekends :)

forabi:
Great! Feel free to work on this!

wholesomedev (commented):
Changes LGTM!

PR #333, opened 2018-03-04
Refactor tests to allow easier overriding of data responses, add tests for NP image, attribution link, search page link

forabi:


wholesomedev (commented):
LGTM

PR #335, opened 2018-03-04
Increase NP image size on desktop

### LINKED ISSUE #322: Make NP image size on NP page match the image size of Related NPs
Currently, on desktop on a widescreen, the Related NP images are bigger than the NP's image at the top. This doesn't look good. I think we should make sure the NP image matches the image sizes of Related NPs, or at least not have Related NP image sizes be bigger than the main NP image size.

forabi:
Closes #322


wholesomedev (commented):
LGTM

PR #340, opened 2018-03-05
Replace coveralls with codecov

forabi:
I'm getting different line positions in coverage reports with coveralls compared to the local ones. I'm wondering if coveralls is to blame here.


forabi:
We'll need to merge this to get a base commit to compare and view the sources in codecov

wholesomedev (commented):
Man, this seems to much better than coveralls. LGTM!

forabi:
Well, it didn't help with #309, but it does look better. Let's stay with codecov!

PR #342, opened 2018-03-05
Upgrade to TS 2.8

forabi:
TypeScript 2.8 has a pretty cool feature, called [Conditional Types](https://github.com/Microsoft/TypeScript/pull/21316), that allows us to do type-level programming, including things like:

* Excluding keys from object types with the new built-in `Exclude<T, U>`
* Unboxing types wrapped in promises like this: `type UnboxPromise<T> = T extends Promise<infer R> ? R : T`
* Getting the return type of a function with the new built-in `ReturnType<T>`.
* Removing `null` and `undefined` from types, with the new built-in `NonNullable<T>`.

`NonNullable` is particularly useful for us because we can get rid of non-null assertions in components that consume API queries that could possibly have a `null` value. See the diff to get an idea of how this could simplify our types.

Not merging this for now because:

* TypeScript 2.8 is not stable yet, will be released this month.
* `typescript-eslint-parser` is failing on the new `infer R` and `T extends R ? ...` syntax.


forabi:
Hmm, I thought I disabled codecov comments :thinking: 

Anyway, here are the built-in conditional types that will be available in TS 2.8: https://github.com/Microsoft/TypeScript/pull/21847

wholesomedev:
Dude, this is awesome and just in time. Yesterday I was googling for "typescript how to remove null from type" and now I have my answer!

forabi:
Yep! Conditional types are very powerful!

wholesomedev (commented):
🎉

PR #346, opened 2018-03-08
Redirect 9 more pages to new Hollowverse

forabi:
Each of the whitelisted pages has around 50 daily unique visits according to Google Analytics


wholesomedev (commented):
I just tested searching some of those pages on Google and landed on the new Hollowverse ✨🚀🎉!

PR #347, opened 2018-03-08
Fix overflowing PersonDetails content

forabi:
Fixes this (when no summary text is available)

screenshot from 2018-03-08 20-32-08

Also fixes horizontal overflow of the moving illumination while loading.

### DIFF
--- a/src/app/components/PersonDetails/PersonDetails.module.scss
+++ b/src/app/components/PersonDetails/PersonDetails.module.scss
@@ -83,6 +83,7 @@
       width: 100%;
       max-width: $page-max-width;
       align-self: center;
+      overflow: hidden;
     }
 
     .photo-link {
@@ -141,6 +142,8 @@
     }
 
     &.is-loading {
+      overflow-x: hidden;
+
       .photo {
         @include photo-border(transparent, transparent);
       }


wholesomedev:
@forabi I don't see any issues with the posted image 🤔

forabi:
The person details section has a scrollbar, separate from the page's
scrollbar.

wholesomedev:
Oh yeah, I see it now. Wouldn't a better fix be to remove this line? https://github.com/hollowverse/hollowverse/pull/347/files#diff-71a8d6762dda41374d8b7e4951a5a427R10

forabi:
> Wouldn't a better fix be to remove this line? 

Nice catch! I'm just wondering how you found out that removing `overflow-x: hidden;` on `.root` fixes the vertical overflowing of content? If I understand [this](https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue/6433475) correctly, setting `overflow-x: hidden` causes the browser to change the default value of `overflow-y` to be `visible`.

wholesomedev:
I wish I had an answer to make me look smart, but the truth is I just felt that an element shouldn't scroll by default. It should simply push the content under it down. So I suspected there might be an unusual overflow setting some where and that `overflow-x` happened to be the first thing I played with (on the live website) and it fixed the issue 😄

PR #356, opened 2018-03-11
Some improvements to NotablePerson and Search pages

wholesomedev:
This PR includes the following changes:

1. Rename `SearchView` to `SearchBar`
1. Move illumination effect to a Sass mixin and share between Search and NotablePerson pages
1. Set different `head` values for the different Notable Person page states
1. Move all `SearchResults` content to inside `Card`. I thought this provided a better UX because it made the transition between the results and no-results state less jarring. It is closer to Facebook Search UX.
1. Remove `FlipMove`. Algolia Search is designed to be fast and to react to every keystroke. The Algolia Search speed with FlipMove animation created a lot of commotion. I felt the Search page worked better without FlipMove. (I also tried debouncing the functions, but then checked official Algolia examples and found that they don't do debouncing, so I reverted those changes)


forabi:
Looks good so far! Are we going to re-implement the skeleton like we did for the NP pages?

wholesomedev:
@forabi Yes, but I wanted to do that as part of another PR when I get the opportunity.

wholesomedev:
I've pushed additional changes to this branch that include experimenting with a new render method naming convention for `page` level components and implementing the search skeleton as part of the SearchResults markup.

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -121,57 +135,43 @@ const Page = withRouter(
                   : commentsUrl
               }
             />
-            <title>{`${name}'s Religion and Political Views`}</title>
```
forabi, line 124: I've deliberately used a template string here because I was sometimes getting an extra whitespace between the name and the `'s`. Not sure why or if it still happens.

forabi (commented):
This is awesome! I just tested the UI changes, I'll do a more in-depth review of the code tomorrow.

src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -106,12 +91,41 @@ const Page = withRouter(
...
+          }
+          icon={warningIcon}
+        />
+      </>
+    );
+
+    renderLoadingStatus = () => (
+      <>
+        <Helmet>
+          <title>Loading notable person...</title>
```
forabi: Do we really need that? Maybe just remove this and let Helmet fallback to "Hollowverse".

wholesomedev: You mean you don't like the cool action I added to the titlebar? 😄

Just kidding. Yeah, I agree, it's not necessary. I'll remove it

src/app/pages/NotablePerson/NotablePerson.tsx
wholesomedev: I'm not sure why that was happening with you. I just tested and things seem fine?

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -58,107 +58,120 @@ const Page = withRouter(
...
+          <Helmet>
+            {userHasPerformedSearch ? (
+              <title>Search results for {searchQuery}</title>
+            ) : (
+              <title>Search for influential people</title>
+            )}
+          </Helmet>
+          {userHasPerformedSearch ? (
+            <ResultsList
+              // why can't TS tell that `value` here is guaranteed to be `AlgoliaResponse`? 🤔
```
forabi, line 119: It's not that smart. In order to get "control flow analysis" to kick in, `value` has to be in an `if` statement or similar.

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -58,107 +58,120 @@ const Page = withRouter(
...
+        <>
+          <Status code={200} />
+          <Helmet>
+            {userHasPerformedSearch ? (
+              <title>Search results for {searchQuery}</title>
+            ) : (
+              <title>Search for influential people</title>
+            )}
+          </Helmet>
+          {userHasPerformedSearch ? (
```
forabi, line 117: If you replace `userHasPerformedSearch` with `!!searchQuery && !!value`, it will know that `value` is of type `AlgoliaResponse`. I think this is a limitation that could be fixed in TS, because `userHasPerformedSearch` is a constant. There is no way its value could have changed between the time it was assigned and the time it was used.

src/app/pages/NotablePerson/NotablePerson.tsx
forabi: I think it was only happening with SSR markup and JS disabled, but I just tried that and there are no issues! :+1:

forabi:
I've updated some of the code in `ResultsList` to use more explicit types to help TS do a better job at type checking. Let me know if this change makes sense. I think we should also add tests for 5xx error states (for both search results and notable person pages). I'll do that later today.

forabi:
Compiling tests with webpack proved to be very slow as we added more tests. I now think we should switch to regular Jest transformers.

forabi (approved):

wholesomedev (commented):
Thanks for adding those fixes and the tests. Looks good! 👍

PR #365, opened 2018-03-19
Switch to regular Jest test setup

forabi:


forabi:
Looks like this is not going to be as easy as I thought. Server rendering is completely dependent on Webpack.

forabi:
Actually, it's pretty easy :smile:. I got it to work. We just need to not use `webpack-flush-chunks` in tests.

forabi:
Hmm, after a lot of tinkering, this is failing because of a dependency conflict between `babel-minify-webpack-plugin` and `ts-jest`. The tests are actually passing but the build command is failing when trying to minify the code. This is only blocked on a version update for either of the tools to switch to Babel 7.

forabi:
After hours of bug hunting, I tracked down the bug to `babel-preset-minify` (so it's not the webpack plugin per se). I have had many issues with this preset before. We can workaround this temporarily by transpiling to ES5 and using UglifyJS instead (which means we are no longer using `@babel/preset-env`).

forabi:
The size difference between the ES5 build and `preset-env` build is insignificant.

screenshot from 2018-03-22 21-01-07
screenshot from 2018-03-22 21-00-45

wholesomedev:
Yeah, doesn't look like a big deal. If it'll allow us to get rid of a dependency that keeps giving us trouble, I think it's totally worth it.

forabi:
Great! Let's merge this and open an issue to switch back to `preset-env` when `preset-minify` becomes stable enough.

PR #367, opened 2018-03-21
Pass all environment variables strating with ENC_PASS_ or TRAVIS_ to build container

forabi:
If we add more secrets in the future, we don't have to remember to add them to the `docker run` command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,8 @@ script:
     -e IS_PULL_REQUEST=$TRAVIS_PULL_REQUEST
     -e AWS_ACCESS_KEY_ID
     -e AWS_SECRET_ACCESS_KEY
-    -e ENC_PASS_SUMO
-    -e TRAVIS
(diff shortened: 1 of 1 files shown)


wholesomedev (commented):
I have no idea how this works, but it looks cool. LGTM!

forabi:
In addition to passing individual environment variables using `-e`, Docker run accepts an environment file using `--env-file` of the following format:

```
VARIABLE_1=VALUE_1
VARIABLE_2=VALUE_2
```

Linux has a command called `env` that prints all the environment variables defined in the shell instance with the same format.

The syntax `--env-file <(env | grep ENC_PASS_)` tells the shell to create something like a virtual file whose contents are all the available environment variables (`env`), run through (`|`) `grep` which filters the input with the regular expression `ENC_PASS_` and returns the matching lines.

wholesomedev:
🤯

PR #371, opened 2018-03-23
Restore code coverage badge

forabi:
Moving to regular Jest setup removed a lot of Webpack code in the test files, which increased our coverage to around 80%. The coverage badge is now yellow. I don't know if we should restore it.


forabi:
@wholesomedev What do you think?

wholesomedev (commented):
As discussed on Discord, we should probably not merge this for now since the badge is yellow even though there are no issues with our test coverage.

PR #372, opened 2018-03-23
Revert "Switch to regular Jest test setup"

forabi:
Reverts hollowverse/hollowverse#365

It seems that UglifyJS is producing minified code that breaks on Safari. When I select a notable person page from the related people list, the page gets stuck loading. The only helpful error message I could get debugging in Safari is: `SyntaxError: cannot create duplicate variable (text trimmed)`.


wholesomedev (commented):
Bummer...

PR #378, opened 2018-03-30
Migrate to AWS Lambda

forabi:


forabi:
The changes in this PR so far are the simplest way to migrate to AWS Lambda. All that's need is to add a new file that wraps the main server as a Lambda function.

The downside is that we won't have a Lambda-like environment during development. We'll continue with the same development setup that we currently have.

I tried using `serverless-webpack` and `serverless-typescript-plugin` but they are both just too hard to integrate with our non-standard Webpack config. `serverless-webpack` is designed to run client-side apps _or_ Node.js apps, not a hybrid SSR SPA.

wholesomedev:
> The downside is that we won't have a Lambda-like environment during development.

Why is this a downside?

forabi:
> Why is this a downside?

Honestly I don't know if it's a good or a bad thing, but the development environment will be very different from AWS Lambda. For example, I managed to get `serverless-offline` to simulate AWS Lambda + API Gateway, and it turns out API Gateway does not support brotli compression and this caused responses to be malformed in Chrome. I had to disable brotli compression to get it to work. This is an example of something that could go unnoticed in development and cause a production issue.

It also feels like we are not really taking advantage of the Serverless framework. It's only used for deploying the production build.

forabi:
Should we continue with this?

wholesomedev:
Yeah, I feel that we should. The incompatibility between Serverless and our
webpack is something that can be ironed out. Having a serverless
architecture is worth it in my opinion.

forabi:
:+1:

clownOverrides/clownCallback.js
```diff
@@ -5,10 +5,17 @@ module.exports = function clownCallback(clownFs) {
...
+  clownFs.remove(babelRcPath);
+
   /* eslint-disable no-param-reassign */
   clownFs.editJson(packageJsonPath, json => {
     delete json.jest;
     delete json.devDependencies['npm-run-all'];
+    delete json.dependencies['@hollowverse/utils'];
 
     return json;
   });
```
wholesomedev: Can you rebase on `master`? This is different in `master` now. I've made `clownFs` expect relative paths, not absolute paths.

wholesomedev: Paths relative to `process.cwd()` that is...

forabi: Done

forabi:
@wholesomedev This one is also ready. Want to take a look? Once we merge this we won't be able to update hollowverse.com until we switch the domain to the new architecture.

I wanted to further split the serve into two functions: `serveSsrMarkup` and `logEvents` but I don't think that has many benefits over a single function. Maybe we can do that in a future PR if it does not complicate the development experience.

I also wanted to upload the static assets to S3 but then I realized that this is not really required to get things working. We can keep the files on Lambda and they will be cached via CloudFront just as if they were on S3.

wholesomedev (approved):
This looks great! 

I'm surprised though that the change is not too big in the source code side?

Also, I'm a little confused about how things work now. I think of an Express server as an always-up Node instance.

But what is an Express server in the context of an AWS Lambda?

For Lambda, do we start a new Express instance for each request and terminate it after the request has been served without keeping any state between requests?

forabi:
> I'm surprised though that the change is not too big in the source code side?

Yep, I didn't expect that. We are basically wrapping the server with `aws-serverless-express` and setting up API Gateway to proxy everything to the function.

> Also, I'm a little confused about how things work now. I think of an Express server as an always-up Node instance.

Express per se is not a server. It's just an object with some functions that takes an HTTP request and returns an HTTP response. The `listen` function on Express creates and starts a regular Node.js server that listens on the specified port. For Lambda, `aws-serverless-express` wraps the Express object and transforms Lambda HTTP events to Express-compatible requests. AWS integration with API Gateway is what triggers the function.

> For Lambda, do we start a new Express instance for each request and terminate it after the request has been served without keeping any state between requests?

Not quite. A single function can process any number of HTTP requests. It's basically a standard Node.js process but it could be terminated by Lambda at any time to free up resources for other AWS customers. The time a Lambda is kept idle before it is terminated is not explicitly documented but [some experiments](https://theburningmonk.com/2017/06/finding-coldstarts-how-long-does-aws-lambda-keep-your-idle-functions-around/) estimate that time to be around 1 hour.

wholesomedev:
>The time a Lambda is kept idle before it is terminated is not explicitly documented but [some experiments](https://theburningmonk.com/2017/06/finding-coldstarts-how-long-does-aws-lambda-keep-your-idle-functions-around/) estimate that time to be around 1 hour.

I don't think the time the Lambda is kept idle is relevant to our discussion here. 

I don't imagine that there is a state that is kept while the Lambda is idle and then lost when the Lambda is terminated? So I guess every Lambda invocation is processed somewhat as a pure function?

forabi:
> I don't imagine that there is a state that is kept while the Lambda is idle and then lost when the Lambda is terminated? So I guess every Lambda invocation is processed somewhat as a pure function?

True. But that was also true for the old server. We chose not to keep any state between requests, but that has nothing to do with being on Lambda vs a regular server. It's just a Node.js process and we can store whatever state we like and it will be kept until the process is terminated.

forabi:
In case I didn't clarify that part: a _single_ Express instance will process _all_ the incoming requests until Lambda terminates the process because it's been idle for a while.

PR #381, opened 2018-04-03
 Use GET for API queries

### LINKED ISSUE #361: Implement caching for API
We'd probably also want to cache API responses since most requests are going to look the same.

forabi:
Eventually, when we implement user log-in, we may need to configure `GraphQLClient` to use `GET` for non-logged in users (to take advantage of CDN cache) and `POST` for logged-in users who can perform mutations (`POST` is required for mutations).

Closes #361


package.json
```diff
@@ -142,8 +142,7 @@
     "empty-module": "^0.0.2",
     "express": "^4.15.2",
     "graphql": "^0.11.0",
-    "graphql-request": "^1.4.0",
-    "graphql-tag": "^2.5.0",
+    "graphql-request": "npm:@forabi/graphql-request@1.6.3",
```
wholesomedev, line 145: What is `npm:`?

wholesomedev (commented):
LGTM!

package.json
forabi: It's the registry that should be used to resolve the package (you can also use `git:` and `file:`). This is a yarn specific thing and is not supported by npm. It's handy for aliasing packages without webpack though. The package would be installed in `node_modules/graphql-request` so packages depending on `graphql-request` would still work.

wholesomedev: Ah, that sounds really handy. Awesome!

PR #382, opened 2018-04-03
Fix API client not sending requests as GET

forabi:
Due to a [silly mistake](https://github.com/graphcool/graphql-request/pull/69/commits/b2a34dddd934039298ee66e98bca3a7688862dc7) in my patched `graphql-request`, specifying `GET` as the `method` did not take effect.


wholesomedev (commented):
LGTM

PR #394, opened 2018-04-08
Use Clown

wholesomedev:

PR #395, opened 2018-04-08
Attempt to fix different date strings by timezone in snapshots

### LINKED ISSUE #393: Jest tests fail when run on a machine with a different timezone?
When I run Jest on my local machine, I get the following error

image

forabi:
The date string returned from the API does not have any timezone info (because the original data didn't). We have to assume a specific timezone to get constant date strings across different timezones.

Closes #393

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "app/test/jest/dev": "jest --color --watchAll",
     "app/test/dev": "npm-run-all --aggregate-output -s clean/test -p 'app/test/*/dev'",
     "app/test/webpack": "NODE_ENV=test webpack --config src/webpack/webpack.config.test.js",
-    "app/test/jest": "jest --coverage --verbose --color",
+    "app/test/jest": "TZ=UTC jest --coverage --verbose --color",
     "app/test": "run-s clean/test app/test/webpack app/test/jest coverage/remap",
     "lint": "run-p -c --aggregate-output lint-css lint-js lint-graphql lint-ts check-ts",
     "lint-css": "stylelint --syntax scss 'src/app/**/*.scss'",


forabi:
@wholesomedev Can you try this change and see if it fixes #393.

wholesomedev:
I pulled `forabi:attempt-to-fix-date-timezone`, `rm -rf node_modules` (just for good measure), then `yarn install`, then `yarn test` but it still failed with the same date issue

forabi:
Hmm, welcome to the JavaScript `Date` mess...

wholesomedev:
@forabi Can we mock the date functions for the tests?

forabi:
> @forabi Can we mock the date functions for the tests?

I don't like mocking in general, and I think setting a time zone on the date is a better solution. I just made some changes to this branch, can you try again now?

forabi:
@wholesomedev The weird thing about this issue is that the tests pass both on my machine and on CI. They are definitely on different timezones. So why does it fail for you? Could it be because of different OSes?

forabi:
> The weird thing about this issue is that the tests pass both on my machine and on CI. 

Well, it's not weird I guess, because CI and my machine could still be on the same day, even if the timezones are different.

wholesomedev:
@forabi I pulled, `yarn install`, `yarn test`, still same issue.

It seems to be failing on the CI too now?

forabi:
It's failing on CI because the bundle size increased.

forabi:
I changed the timezone on my machine to match yours and I was able reproduce the issue.

The solution is very simple and elegant -- it turns out we can set the timezone for Jest with the `TZ` env variable:

```
TZ=UTC jest
```

That did the trick!

forabi:
@wholesomedev Feel free to merge this after confirming it works for you. We should rebase #394 on master after merging this and revert the change to the snapshots.

wholesomedev:
This works! 🎉

wholesomedev (approved):

PR #404, opened 2018-04-12
Restore PROJECT env variable in build container

forabi:
This is still needed for Elastic Beanstalk projects


wholesomedev (commented):
LGTM

Issue #389, opened 2018-04-07
Allow Clown CLI to execute a callback function per extension folder


wholesomedev:
This would give a lot of flexibility to extension folders to allow them to do some post-processing.

The main problem with this approach is that it would break the `clown check` functionality. The way to fix this would be to make clown check use a fully mocked file system, which is something that has not worked very well for me when I tried it.

forabi:
I was going to sleep, but I just had this weird idea and I couldn't help but share it :smile: 

What if we use [Node.js `vm` module](https://devdocs.io/node/vm) to create an isolated JavaScript context to execute the callback function? That way we can limit what the function can do and even provide our own utility functions to be used in `clown.js`.

Here is a basic implementation:

```typescript
const vm = require('vm');
const fs = require('fs');
const sandbox = {
  deleteFile: async (path) => { await fs.unlink(path); }
};
vm.createContext(sandbox);

const code = '<read clown.js as a string>';
vm.runInContext(code, sandbox);

```

We could easily mock `deleteFile` for dry runs (i.e. `clown check`).

forabi:
The callback code could be as simple as:

```js
deleteFile('/path/to/file');
```

wholesomedev:
@forabi what's the benefit of using `vm` here? Wouldn't this be the same as passing a `sandbox` object as an argument to the `clown.js` callback function?

forabi:
> Wouldn't this be the same as passing a sandbox object as an argument to the clown.js callback function?

If we just pass this object, the callback code and still do this:

```js
const fs = require('fs');

fs.unlink('path/to/file');

```

So the "sandbox" is not really doing anything, it can easily be escaped. With `runInContext`, `require` is not even defined in the function's context. Think of it like how workers cannot access the DOM in the browser. It's an entirely different environment, the only thing in common is that it uses JavaScript as the language.

wholesomedev:
@forabi Oh, I see. That's interesting. 

Yeah, this could be a viable solution. It would limit what the callback functions can do though, maybe too much. 

Passing arguments to the callback functions gives responsibility to the callback function, which is not ideal either.

If we can get a mocked filesystem to work properly during dry-runs, that would be ideal I think.

forabi:
> It would limit what the callback functions can do though, maybe too much

Yes, that's true.

wholesomedev:
I got an idea about this one. I should've thought of this earlier! 😅

[...]

PR #405, opened 2018-04-12
Add npm-run-all to production dependencies

forabi:
This actually caused deployment to fail as the start command depends on `run-p`

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,7 @@
     "lru-cache": "^4.1.1",
     "moment": "^2.19.1",
     "node-fetch": "^1.7.3",
+    "npm-run-all": "^4.0.2",
     "path-to-regexp": "^2.1.0",
     "preact": "^8.2.7",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -317,7 +317,8 @@
     "webpack-merge": "^4.1.1",
     "webpack-node-externals": "^1.6.0",
     "wildcards-entry-webpack-plugin": "^2.1.0",
-    "yarnhook": "^0.1.1"
+    "yarnhook": "^0.1.1",
+    "npm-run-all": "^4.0.2"
```
wholesomedev, line 321: 🤡

wholesomedev (commented):
LGTM

package.json
forabi: Ideally we should remove this from `devDependencies` but clown would complain.

wholesomedev: Yeah. #389 would help solve this problem

PR #406, opened 2018-04-12
Switch to regular Jest setup (attempt 2)

forabi:
I'm happy to report that minification with UglifyJS no longer breaks code on Safari. A few package upgrades and configuration changes did the trick.


wholesomedev (commented):
Great that we were able to simplify this piece!

This LGTM! 🎉

PR #407, opened 2018-04-13
Upgrade to config v10

forabi:


forabi:
Looks like `yarn` will rewrite `package.json` to remove duplicate packages in "dependencies" and "devDependencies". Merging this will cause deployment to fail (same as #405).

wholesomedev:
@forabi should we move `npm-run-all` to `dependencies` in `clown/common` for now?

forabi:
@wholesomedev Never mind, I quickly looked at the diff and thought `npm-run-all` was removed from `dependencies`, but the entries were just re-sorted in `devDependencies`.

wholesomedev (commented):
LGTM

PR #408, opened 2018-04-13
Use objects for dev helpers params, use @hollowverse/utils/helpers/env

forabi:
## Use objects as parameters to some dev helpers:
  * Code is easier to understand, we don't have to look at the implementation to know what this parameter does
  * We can add as many options as we want without having to worry about default values or intermediate parameters

## Use env helpers from @hollowverse/utils

Depends on #342 (because @holloverse/utils requires TS 2.8)


wholesomedev (commented):
LGTM!

PR #409, opened 2018-04-13
Add promiseToCancelable helper, use it for AsyncComponent

forabi:
React throws an error when trying to `setState` on a component that has been unmounted. This is the case for `AsynComponent` when the `load` function completes after the async component has been removed from the tree. If the user navigates away from an `Image` that has started loading, the component will be unmounted by the time the load has finished.

Inspired by [this article](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html) on the React blog, I've added a new `promiseToCancelable` function that gets rid of this error in `AsyncComponent`. It uses Node.js built-in `EventEmitter` (that also works in browser) to cancel the promise. Ideally we should use the browser's new standard [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget) [but that just does not work in Jest](https://github.com/jsdom/jsdom/issues/2173#issuecomment-373509419).

Before:

screenshot from 2018-04-13 21-14-18

After:

screenshot from 2018-04-13 21-22-11

### DIFF
--- a/src/app/__tests__/notablePersonPage.ts
+++ b/src/app/__tests__/notablePersonPage.ts
@@ -9,6 +9,9 @@ import {
 } from 'fixtures/notablePersonQuery';
 import { EditorialSummary } from 'components/EditorialSummary/EditorialSummary';
 
+const emptyBase64EncodedImage =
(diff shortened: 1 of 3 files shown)


wholesomedev (commented):
I've been looking for an elegant way to cancel promises. This looks sweet!

LGTM!

PR #411, opened 2018-04-15
Add meta description tags to NP and Search pages

### LINKED ISSUE #357: Add meta description tag to Hollowverse pages
We should add meta description tag to the Hollowverse pages, such as Notable Person, Search, and Homepage.

wholesomedev:
Closes #357


wholesomedev:
Not sure why this is failing. Will look into it

src/app/pages/SearchResults/SearchResults.tsx
```diff
@@ -107,9 +108,27 @@ const Page = withRouter(
           <Status code={200} />
           <Helmet>
             {userHasPerformedSearch ? (
-              <title>Search results for {searchQuery}</title>
+              <>
```
forabi: @wholesomedev It looks like Helmet does not support fragments as children? Here is the first line of the error stack on CI:

```
Helmet.warnOnInvalidChildren (/repo/node_modules/react-helmet-async/lib/index.js:169:297)
```

wholesomedev: @forabi Yeah, looks like it. The least resistant alternative way I found to do this is to use the `{condition && <element/>}` pattern for each element. Are you okay with that?

forabi: @wholesomedev It may support arrays:

https://github.com/nfl/react-helmet/blob/master/src/Helmet.js#L36

wholesomedev: Yep, that worked 👍

forabi (approved):
LGTM!

PR #421, opened 2018-04-19
Proxy legacy pages to live.hollowverse.com

forabi:
For some reason proxying to static.legacy.hollowverse.com is not working on Lambda. CloudFront checks the `Host` header of the request and returns 403 if the host is not listed in the CF distribution CNAMEs list. `changeOrigin` in proxy settings should set the `Host` header to match the target URL host so CF _should_ accept the request but it is not 🤷.


wholesomedev (commented):
LGTM

PR #425, opened 2018-04-21
Drop SSR

### LINKED ISSUE #250: Try to unify index.html and index.server.html

### LINKED ISSUE #253: Log events are not being sent server-side

### LINKED ISSUE #255: Should we move server-side functionality of logging to an AWS Lambda?
In the spirit of having more of a micro-service architecture, I'm wondering if it would make sense for us to move the logging functionality that we've implemented in `server.ts` to a Lambda.

### LINKED ISSUE #332: Show server-side rendering errors in browser during development
I think it would be helpful to show SSR errors some where in the browser during development because as I’m doing front end work, I don’t think to look for errors in my terminal window. I expect to see the errors in the browser.

(I’m just trying to think of ways to make SSR less impactful to developer ergonomics)

### LINKED ISSUE #339: Hot Loader errors in console
Sometimes I these errors in console

image

Can we do something about these errors? Can we configure Hot Loader to reload the page when it fails to hot reload content?

### LINKED ISSUE #427: Drop SSR and simplify build processes
Based on discussion at #424

forabi:
Fixes #427 
Fixes #339 
Closes #250 
Closes #253
Closes #358 
Closes #332 
Closes #255 

To do:
* [x] Re-implement log endpoint as a separate Lambda, or use it in `appServer`.


wholesomedev (commented):
This was a really valuable experiment and we learned a ton from it.

I like the way the code now looks.

LGTM!

PR #426, opened 2018-04-22
Fix loading flash

### LINKED ISSUE #323: "Loading..." text flashes on non NP pages

wholesomedev:
I've used our loading spinner as a substitute for the "Loading..." flash. I changed the CSS of the loading spinner to make it always vertically and horizontally centered within its parent.

Fixes #323

### DIFF
--- a/src/app/__tests__/notablePersonPage.ts
+++ b/src/app/__tests__/notablePersonPage.ts
@@ -2,7 +2,6 @@ import {
   ClientSideTestContext,
   createClientSideTestContext,
 } from 'helpers/testHelpers';
-import { delay } from 'helpers/delay';
 import { pageLoadSucceeded } from 'store/features/logging/actions';
 import {
   notablePersonWithEditorialSummaryQueryResponse,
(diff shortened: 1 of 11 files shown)


src/app/__tests__/notablePersonPage.ts
```diff
@@ -25,7 +26,11 @@ describe('Notable Person page', () => {
       });
     });
 
-    it('has notable person name', () => {
+    it('has notable person name', async () => {
+      // We force a 300 ms min delay in our loadable components for better UX. The delay below ensures
+      // that the loadable component has rendered before making the assertion.
+      await delay(301);
```
wholesomedev: I'm not sure how to better handle this case

forabi: I think we should go back to using a custom routes map for tests which uses the actual, resolved components instead of the loadable ones. Anyway, this needs to be rebased as I've switched to using `react-loadable` because `react-universal-components` is for SSR apps and it didn't work well with Webpack 4 (it could have worked client-side only with Webpack 3). I'll rebase and make the necessary changes.

forabi (approved):

wholesomedev (commented):
Awesome, thanks!

PR #429, opened 2018-04-24
Ensure cancel listener is removed on `Cancelable`

forabi:
This avoids a possible memory leak where the listener does not get removed because the `cancel` event was never dispatched.

### DIFF
--- a/src/app/helpers/promiseToCancelable.ts
+++ b/src/app/helpers/promiseToCancelable.ts
@@ -15,14 +15,25 @@ export function promiseToCancelable<T>(promise: Promise<T>): Cancelable<T> {
   let wasCanceled = false;
 
   const target = new EventEmitter();
+
   const cancelationPromise = new Promise<T>((_, reject) => {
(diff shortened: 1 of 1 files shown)


src/app/helpers/promiseToCancelable.ts
```diff
@@ -15,14 +15,25 @@ export function promiseToCancelable<T>(promise: Promise<T>): Cancelable<T> {
...
+    .then(value => {
+      target.removeAllListeners();
+
+      return value;
+    })
+    .catch(error => {
+      target.removeAllListeners();
+
+      throw error;
+    });
```
wholesomedev, line 36: I wonder if we have/can use `finally` here? 🤔

wholesomedev (commented):
LGTM

src/app/helpers/promiseToCancelable.ts
forabi: Right! I forgot about `finally`. I'll sneak the change in #435

PR #430, opened 2018-04-24
Refactor AsyncResult types and related components

### LINKED ISSUE #484: Log events of type "PAGE_LOAD_FAILED" should include error details if possible

forabi:
Refactor `AsyncResult`, which is a ["discriminated union type"](http://www.typescriptlang.org/docs/handbook/advanced-types.html) to help the compiler infer types better.

Also introduce a new kind of `AsyncResult`: `StaleResult` which replaces what used to be `OptimisticResult`.

Re-use `OptimisticResult` to allow deep partial data respones.

Thanks to conditional types, we can now have a deep partial object that can be accessed safely.

This is useful for example for taking the notable person name and image from the search results and re-using them in `NotablePersonBody` while the full API query is being fetched.

Also be more strict with the parameter type of `renderNonErrorStatus`: it no longer silently accepts `ErrorResult` types.

Fixes #484

### DIFF
--- a/src/app/pages/NotablePerson/NotablePerson.tsx
+++ b/src/app/pages/NotablePerson/NotablePerson.tsx
@@ -95,9 +95,9 @@ export const NotablePerson = withRouter(
       );
     };
 
-    renderErrorStatus = () => (
+    renderErrorStatus = (error?: Error) => (
       <>
-        <Status code={500} />
+        <Status code={500} error={error ? error.message : undefined} />
         <Helmet>
           <title>Error loading notable person page</title>
         </Helmet>
@@ -191,7 +191,7 @@ export const NotablePerson = withRouter(
               {({ result }: { result: Result }) => (
                 <div className={classes.root}>
                   {isErrorResult(result)
-                    ? this.renderErrorStatus()
+                    ? this.renderErrorStatus(result.error)
                     : this.renderNonErrorStatus(result)}
                 </div>
               )}
--- a/src/app/pages/NotablePerson/NotablePerson.tsx
+++ b/src/app/pages/NotablePerson/NotablePerson.tsx
@@ -5,8 +5,9 @@ import { oneLine } from 'common-tags';
 
 import {
   isErrorResult,
-  isPendingResult,
   AsyncResult,
+  isSuccessResult,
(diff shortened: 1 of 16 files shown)


src/app/pages/NotablePerson/NotablePerson.tsx
```diff
@@ -94,9 +95,9 @@ const Page = withRouter(
       );
     };
 
-    renderErrorStatus = () => (
+    renderErrorStatus = (error?: Error) => (
       <>
-        <Status code={500} />
+        <Status code={500} error={error ? error.message : undefined} />
```
wholesomedev, line 100: Since we're not doing server side rendering anymore, does this `Status` component still make sense?

We should probably handle 404 in our web server eventually...

forabi: We need a way to indicate global page load results so that they can be logged. I know `Status` may not make much sense. Maybe we could rename it to `PageLoadStatus` and replace the `code` prop with a `state` prop with values like `status` and `error`? What do you think?

forabi: > We should probably handle 404 in our web server eventually...

Now that we dropped SSR, I'm wondering how we would do that. We could intercept the requested path, say '/Tom_Hanks' and check with the API that this NP exists, but that would only work for NP pages. What about search? We would need to do the search on the server as well. It feels like we are doing SSR again :smile:

wholesomedev: Hmm, interesting. It didn't occur to me how these statuses could get tricky without server-side code!

>We need a way to indicate global page load results so that they can be logged. I know `Status` may not make much sense. Maybe we could rename it to `PageLoadStatus` and replace the `code` prop with a `state` prop with values like `status` and `error`? What do you think?

Not sure. It seems to me like a low priority thing at the moment. Should we log an issue for it and revisit it later?

forabi: Yep, no problem. Let's merge this and open an issue for `<Status />`

wholesomedev (commented):
LGTM

PR #431, opened 2018-04-24
Work around NavBar issue on Safari for iOS

forabi:
Force the content `position` to `absolute` and set the `top` to the height of the navigation bar. This is hacky and somewhat breaks the encapsulation of the `Sticky` component.

Fixes #428

PR #433, opened 2018-04-24
Revert "Work around NavBar issue on Safari for iOS (#431)"

forabi:
This reverts commit 11bec03006c0315ef9022da9f61d31c03c2017bc.

While this workaround did work, it prevented pages from having explitic heights so we could
no longer have the search results list take the whole height of the page even if empty.


wholesomedev (commented):
LGTM

PR #434, opened 2018-04-24
Try to fix NotablePersonBody jumping while loading

forabi:
Observations:

* It happens on all browsers, even on desktop, so it's not a Safari bug
* It happens only once when an NP page is loading for the first time. So it's most likely in `LodableNotablePerson`
* Looking at the structure of `loading` component in `LoadableNotablePerson` we see that it's not exactly the same as the fully loaded page:

```tsx
// loading
() => <NotablePersonBody />

// fully loaded
() => (
 <div classeName={classes.root}>
   <NotablePersonBody />
   {/* other components */}
 </div>
)
```

So I suspect it's being caused by some missing styles not being applied at the root of the loading component. We need to share some CSS rules between `LoadableNotablePerson` and `NotablePerson` and match the structure by adding a wrapper `<div />`.

It's very hard to test this in development because the styles are applied on the fly by `style-loader` and that causes a flash of unstyled text.

### DIFF
--- a/src/app/pages/NotablePerson/LoadableNotablePerson.module.scss
+++ b/src/app/pages/NotablePerson/LoadableNotablePerson.module.scss
@@ -1,3 +1,13 @@
+@import 'variables.scss';
+
 .root {
-  composes: root from './NotablePerson.module.scss';
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding-bottom: $spacing-20px;
+
+  > * {
+    width: 100%;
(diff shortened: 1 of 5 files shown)


wholesomedev (commented):
Maybe this fixed the issue, because I can't see it in production anymore?

LGTM!

forabi:
> Maybe this fixed the issue, because I can't see it in production anymore?

It does not seem to be fixed, at least on Android :cry:

PR #439, opened 2018-04-26
Use a placeholder image for NPs without photos

### LINKED ISSUE #364: Handle UX of NP without image in search results list

forabi:
## Search page
screen shot 2018-04-26 at 21 24 47

## NP page
screen shot 2018-04-26 at 21 24 36

Fixes #364


forabi:
@wholesomedev Do you think this looks good? I tried to choose an image that's gender-neutral. I picked a free one from https://www.flaticon.com/search/3?word=people (free usage requires attribution)

wholesomedev (approved):
Yeah, I think this looks good. How are you going to attribute it though? Or do we need to pay for it?

forabi:
Well, it turns out we either have to attribute the icon in the footer of the page or pay a recurring payment for accessing the entire icon collection. We just need this one icon so I hoped it would be a one-time payment but that option is not available.

I've replaced the icon with one from https://material.io/icons to avoid any possible licensing issues.

Their homepage states:

> The icons are available under the Apache License Version 2.0. We'd love attribution in your app's "about" screen, but it's not required.

Here's what the icon looks like:

screen shot 2018-04-27 at 16 52 44
screen shot 2018-04-27 at 16 52 41

@wholesomedev Let me know what you think.

wholesomedev:
Is there a more central place where we can add this placeholder image to NPs who are missing a photo?

A central place would be better because then we wouldn't have to always remember to handle the case of the missing photo. For example, there's also the Related People section where an NP could have a missing photo.

forabi:
Maybe we should create a new `NpPhoto` component. What do you think?

wholesomedev:
@forabi Yeah, I think creating a NotablePersonPhoto component might be better

forabi:
I created the new component, `NotablePersonPhoto` and updated the code to use it in all the places where we have NP photos.

screen shot 2018-04-28 at 16 52 53

wholesomedev (approved):
Great! I think this is a good implementation. Still not 100% sure about the UX, but this is definitely an improvement on what we had. We can run this by Max at some point in the future if we want to improve the UX. 👍

forabi:
> Still not 100% sure about the UX, but this is definitely an improvement on what we had.

Agreed! We need a better UX. The placeholder looks cartoonish.

PR #441, opened 2018-04-28
Change how Sticky works

### LINKED ISSUE #428: NP page is not scrolled to the top when it first loads
Steps to reproduce:

* Go to https://dev.hollowverse.com/Ray_J on Safari iOS

Expected behavior: NP image is not covered by the page's nav bar

Actual behavior: NP image is partly covered by the page's nav bar. Scrolling a little bit to the top is needed to view the full image.

Does not affect Chrome on Android.

forabi:
Following discussions in #428, this PR changes the behavior of the sticky navbar at the top to match what's described below:

> For example: we give the navbar position static (default positioning), and only when it has completely scrolled out of view, we change its positioning to fixed and animate it down.

I don't know if that fixes #428, I'm going to deploy this to `internal` branch and test the changes.

### DIFF
--- a/src/app/components/Sticky/Sticky.module.scss
+++ b/src/app/components/Sticky/Sticky.module.scss
@@ -1,5 +1,16 @@
+@keyframes slide-down {
+  0% {
+    transform: translateY(-100px);
+  }
+
+  100% {
+    transform: translateY(0);
+  }
+}
+
 .container {
   position: relative;
+  width: 100%;
 
   .element,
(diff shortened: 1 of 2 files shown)


forabi:
The changes are up on https://hollowverse.com/Tom_Hanks?branch=internal

Sadly, this does not seem to fix #428.

forabi:
Closing because it didn't help fix #428 and it caused some jank on Android.

PR #442, opened 2018-04-28
Wrap scroll call in requestAnimationFrame

forabi:
I suspect that Safari is throttling/skipping some calls to `scrollIntoView` which is leading to the issue described in #428.


wholesomedev (commented):
LGTM!!

PR #447, opened 2018-05-01
Do not cache index.html

forabi:
This file contains references to static files that may have been expired in the edge cache and would never be found when requested from the browser. I discovered this while browsing hollowverse.com and I got an empty page. Looking at the network tab in dev tools, I noticed requests to some scripts are failing with 404.


wholesomedev (commented):
👍

PR #450, opened 2018-05-03
Fix SumoLogic logging

### LINKED ISSUE #449: Logs are not being sent to Sumo Logic
I was working on #280 and I've just realized that there has been no logs in Sumo Logic for at least 30 days. This is strange because the log request does not fail in the browser:

screenshot from 2018-05-03 21-18-18

and invoking the function manually does not fail either:

screenshot from 2018-05-03 21-18-01

I've checked the collector ID in the function's environment variables and it is correctly set. There's no errors logged in the manual execution output, so `onError` is not being called on Sumo Logic `logger` instance [as defined here](https://github.com/hollowverse/hollowverse/blob/fa81a9ddb39fa908f9eb8fe4524e46db486134fd/src/logger/logger.ts#L24).

forabi:
Initially I thought that logging was failing because I was using the wrong env variable name (`COLLECTOR_ID` instead of `SUMO_COLLECTOR_ID`), but fixing this didn't fix the issue. I ran a "live tail" on Sumo Logic to monitor logs in real time and could not see any events being logged. This suggested that the actual cause is that Sumo Logger is not flushing logs. The official JS library batches logging requests but does not provide a way to wait until logs are sent, so I've switched to a simple `POST` request and we now `await` for the logs to be sent and I immediately saw the logs in the live tail!

Fixes #449


wholesomedev (commented):
LGTM

PR #452, opened 2018-05-05
Use CSSTree Validator to "type check" CSS

forabi:
(see https://github.com/hollowverse/config/pull/83)


src/app/pages/NotablePerson/RelatedPeople.module.scss
```diff
@@ -17,7 +17,6 @@
   -webkit-overflow-scrolling: touch;
 
   overflow-x: scroll;
-  overflow-x: overlay;
```
forabi, line 20: This is a non-standard property, only supported in WebKit and it seems to have no effect in recent versions of Safari.

src/app/components/TextPage/TextPage.module.scss
```diff
@@ -53,7 +53,6 @@
   a {
     color: $color-blue;
     text-decoration: none;
-    text-decoration-skip: ink;
```
forabi, line 56: This is already set in global CSS.

src/app/components/NavBar/NavBar.module.scss
```diff
@@ -7,7 +7,6 @@
       background: linear-gradient(50deg, #4cfde9 0%, #fbc954 50%, #4435f7 120%);
       // stylelint-disable-next-line property-no-vendor-prefix
       -webkit-background-clip: text;
-      background-clip: text;
```
forabi, line 10: Non-standard property, unprefixed version is not supported by any browser.

src/app/components/App/App.module.scss
```diff
@@ -38,7 +38,7 @@
   }
 
   a {
-    text-decoration-skip: ink;
```
forabi, line 41: The property name has been changed in the spec.

wholesomedev (commented):
LGTM

PR #453, opened 2018-05-05
Simplify log format for SumoLogic

forabi:
The new format is based on the one suggested in ["What is the optimal log format to use with Sumo Logic?"](https://help.sumologic.com/Start-Here/01About-Sumo-Logic/What-is-the-optimal-log-format-to-use-with-Sumo-Logic). I'm trying to make SumoLogic infer most of the key value pairs (https://help.sumologic.com/Search/Search-Query-Language/01-Parse-Operators/04-Parse-Keyvalue-Formatted-Logs) and that should hopefully reduce the need to manually parse or manipulate the logs in SumoLogic dashboard.

I've also created a new collector so that SumoLogic does not have to deal with different log formats.


forabi:
We'll need to test how Sumo Logic infers the key/value pairs with spaces in the values (for things like error messages)

forabi:
It looks like SumoLogic does not play well with spaces in the value. We'll have to parse the key value pairs with regex:

```
_sourceCategory=hollowverse.com | keyvalue regex "\s\[(.*?)=(.*?)\]" keys "type", "payload.message", "payload.source", "payload.line", "payload.column" | where type = "UNHANDLED_ERROR_THROWN"
```

Still an improvement over parsing JSON.

forabi:
@wholesomedev I've shared a couple dashboards with you on SumoLogic, one of which is for unhandled errors using the new format: https://service.us2.sumologic.com/ui/bento.html#/search/PIZh4cODsC6IjNy5FCcxZCmjajYNhNWO77fFHrr7

We can save and share dashboards like this so we don't have to write the parse query every time we want to get information about errors or other log events.

PR #456, opened 2018-05-06
Lint against unused and undefined CSS classes

forabi:
I've been thinking about writing a TSLint plugin to do exactly this. Luckily I've found an ESLint plugin.

What this plugin does:

* Report any CSS Modules classes that are defined in CSS but not used in JSX/TSX files
* Report any CSS Modules classes that are used in JSX/TSX files but are not defined in CSS.

This combined with #452 should significantly improve the CSS development experience. It's like TypeScript for CSS :smile:

screenshot from 2018-05-06 14-18-20

screenshot from 2018-05-06 14-11-43

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": [
     "./node_modules/@hollowverse/config/eslint/base/.eslintrc.json",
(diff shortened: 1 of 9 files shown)


wholesomedev (commented):
Sweet!

PR #462, opened 2018-05-11
Add session ID to log events

### LINKED ISSUE #461: Add a unique user identifier with logged events
This is going to help us calculate metrics such as: pages viewed per user

forabi:
Closes #461

### DIFF
--- a/src/app/helpers/sendLogs.ts
+++ b/src/app/helpers/sendLogs.ts
@@ -1,6 +1,29 @@
 import { Action } from 'store/types';
+import Cookie from 'js-cookie';
+import uuid from 'uuid/v4';
+import { once } from 'lodash';
 
 const logEndpointUrl = `/log?branch=${__BRANCH__}`;
+
+/**
+ * We try to use the Google Analytics "Client ID" to associate
+ * logs with the same GA user, otherwise we generate a new random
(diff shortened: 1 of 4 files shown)


src/app/helpers/sendLogs.ts
```diff
@@ -1,7 +1,30 @@
...
+/**
+ * We try to use the Google Analytics "Client ID" to associate
+ * logs with the same GA user, otherwise we generate a new random
+ * ID.
+ * Note: this function should not be invoked immediately to give
+ * enough time for Google Analytics script to load (since it's loaded lazily).
+ * @see https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id
+ */
+const getSessionId = once(() => {
+  const existingId = localStorage.getItem('uid') || Cookie.get('_ga');
```
wholesomedev, line 17: As we discussed on Discord, I think it would be better if we don't rely on Google Analytics here

wholesomedev (commented):
Looks good, just one comment about using GA cookie.

src/app/helpers/sendLogs.ts
forabi: Addressed in #465

PR #465, opened 2018-05-11
Do not use Google Analytics ID as session ID

forabi:
[It does not seem possible to query GA by session ID](https://developers.google.com/analytics/devguides/reporting/core/dimsmets#cats=traffic_sources,session,user) so this is not useful.


wholesomedev (commented):
LGTM

PR #469, opened 2018-05-12
Invalidate CloudFront on each deployment

### LINKED ISSUE #470: Cache index.html
We need to cache index.html in CF.

Currently the problem with caching index.html is that CF might remove static files referenced within index.html before removing index.html itself. So CF would serve an old index.html which references static files that no longer exist in either CF or the origin server.

We need to either make sure that the origin server can serve any version of static assets or make sure not to serve outdated index.html files.

forabi:
Closes #470.


wholesomedev (approved):
Nice!! 🚀

PR #471, opened 2018-05-12
Update CloudFront invalidate plugin

forabi:
The latest version on npm has a bug which makes the plugin ineffective. We can use the GitHub repo in the meantime.


wholesomedev (approved):
LGTM

PR #480, opened 2018-05-16
Redirect 15 more pages each with around 50 daily unique visits

forabi:
Looks like I was looking at the data for 1 week instead of 1 day. This should get us to 1000 users/day.

### DIFF
--- a/src/app/redirectionMap.ts
+++ b/src/app/redirectionMap.ts
@@ -22,6 +22,21 @@ export const redirectionMap = new Map<string, string>([
   ['/cristiano-ronaldo', '/Cristiano_Ronaldo'],
   ['/bruce-willis', '/Bruce_Willis'],
   ['/brad-pitt', '/Brad_Pitt'],
+  ['/bella-thorne', '/Bella_Thorne'],
+  ['/al-pacino', '/Al_Pacino'],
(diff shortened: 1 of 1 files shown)


wholesomedev (commented):
LGTM

Not sure if this is enough or not. Will keep an eye on the numbers for the next couple of days and post updates if necessary.

forabi:
Stupid question: is there a straightforward way to get unique _users_ per page on GA? I can't seem to find this.

wholesomedev:
@forabi Hmm, not sure. Does [this](https://analytics.google.com/analytics/web/#/report/content-pages/a5715214w55386753p56394045/_u.date00=20180517&_u.date01=20180517&_u.date10=&_u.date11=&explorer-table-dataTable.sortColumnName=analytics.uniquePageviews&explorer-table-dataTable.sortDescending=false&explorer-table.plotKeys=%5B%5D&explorer-table.rowCount=5000/) help?

forabi:
> @forabi Hmm, not sure. Does this help?

Those represent unique pageviews, which is the metric I used to pick the new pages. Unique pageviews do not necessarily correspond to unique users (although in our case, the two metrics might be close).

PR #502, opened 2018-05-26
Revert "Use babel-minify instead of UglifyJS"

forabi:
Well, this broke the website, even though I tested the change on an internal branch :man_shrugging: 

Reverts hollowverse/hollowverse#501


wholesomedev (commented):
LGTM

PR #503, opened 2018-05-26
Use preset-es2015 instead of preset-env

forabi:
Safari is why we can't have nice things

### DIFF
--- a/package.json
+++ b/package.json
@@ -309,7 +309,6 @@
     "tslint-react": "^3.5.1",
     "typescript": "^2.8.1",
     "typescript-eslint-parser": "^15.0.0",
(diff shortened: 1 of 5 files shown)


forabi:
I tested this on an internal branch. Let's see if this reduces the error rate.

wholesomedev (commented):
>Safari is why we can't have nice things

Sounds about right to me 😄

I wonder what the impact of this change could be in terms of performance? 🤔

forabi:
> I wonder what the impact of this change could be in terms of performance? 

Well, considering we were transpiling for Safari iOS 9.3, I don't think it would make much of a difference. It's more of an implementation detail currently.

Issue #505, opened 2018-05-27
Log events of type "PAGE_LOAD_FAILED" should also include the error stack if possible


forabi:
#430 added error messages to the log events, but unfortunately, an error message like `undefined is not an object` is often not useful without the error stack. We should also include the error stack in the log event, and re-map the error position to the source file location on the server, just like we do for `UNHANDLED_ERROR_THROWN`.

PR #506, opened 2018-05-27
Log error stack for failure events

forabi:
This partially (fixes) #505. We still need some work on the server side to parse the error stacks and map the error locations.


wholesomedev (commented):
LGTM

PR #507, opened 2018-05-27
Remove unused fetch polyfill & unused serialize-javascript dependency, refactor RxJS code

forabi:
* Looks like we are no longer using the fetch polyfill. `graphql-request` already loads one so this one is not needed currently.
* `merge` as an operator is deprecated in favor of static `merge`.


wholesomedev (commented):
LGTM

PR #513, opened 2018-05-30
Switch back to using preset-env, target Chrome 41

### LINKED ISSUE #512: New pages not being indexed properly by Google
I was checking our [Google Analytics dashboard](https://datastudio.google.com/u/0/reporting/1sg9ER2pF74eQv0GOLjPLYA5xbSc24eiG/page/1M) to see the number and I noticed that the New Hollowverse "Users" numbers dropped somewhat suspiciously. That could be because of fluctuation in the interest in the NPs who were migrated to the New Hollowverse, or it could be something else...

I thought it might also be that Google is lowering the ranking of the new pages because it can't see their content. So I used the [Fetch as Google](https://support.google.com/webmasters/answer/6066468?hl=en) tool to check how Google sees [Tom Hanks](https://hollowverse.com/Tom_Hanks) page. Unfortunately, I found out that Google only sees a blank gray page without any content!

I did a little bit of research into the matter and found [this article](https://medium.com/@gajus/react-application-seen-as-a-blank-page-via-fetch-as-google-afb11dff8562) which seems to be helpful. And I also saw this [relevant video](https://www.youtube.com/watch?v=JlP5rBynK3E) by a Google Search engineer about how to optimize SPAs for search engines.

My guess here is that we might need to transpile and perhaps polyfill more aggressively because the Googlebot can't execute our JavaScript. Not sure though.

forabi:
Tested on internal branch:

## Before
screenshot from 2018-05-30 08-18-07

## After
screenshot from 2018-05-30 08-24-38

## Fetch as Google
screenshot from 2018-05-30 08-29-29

Should fix #512

### DIFF
--- a/package.json
+++ b/package.json
@@ -307,6 +307,7 @@
     "tslint-react": "^3.5.1",
     "typescript": "^2.8.1",
     "typescript-eslint-parser": "^15.0.0",
+    "uglify-es": "^3.3.9",
     "uglify-js": "^2.4.16",
     "uglify-loader": "^2.0.0",
     "uglifyjs-webpack-plugin": "^1.2.4",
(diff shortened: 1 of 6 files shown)


wholesomedev (commented):
Nice, thanks for taking care of this promptly!

PR #515, opened 2018-05-30
 Refactor log batch payloads

forabi:
* Add timestamp to each action as soon as it is dispatched instead of deferring until the log batch is about to be sent
* Only add `sessionId` and `userAgent` info once per batch. This should save some bytes
* Change signature of `sendLogs` to accept `LogBatch<Action>`, replace with mock implementation in development to make debugging easier

Changes should be transparent to Splunk, i.e. we do not have to update our search queries.


wholesomedev (commented):
LGTM

PR #519, opened 2018-05-31
Fix unhandled rejections

forabi:
This should bring the error rate down as most of the errors are likely CancelationErrors.

It looks like the Promise polyfill provided by core-js (which is added by @babel/preset-env) does not support this way of handling rejections:

```ts
window.addEventListener('unhandledrejection', /* ... */)
```

and only works with:

```ts
window.onunhandledrejection = /* ... */;
```

I can confirm this after looking [into the code](https://github.com/zloirock/core-js/blob/de5ed6f6d00f5bac314aa804262217da85a3e88d/modules/es6.promise.js#L100).

### DIFF
--- a/src/app/index.tsx
+++ b/src/app/index.tsx
@@ -68,7 +68,7 @@ if (module.hot) {
 }
 
 Promise.all([loadIntersectionObserverPolyfill(), loadUrlPolyfill()])
-  .finally(renderOnDomReady)
+  .then(renderOnDomReady)
   .catch(renderOnDomReady);
 
 window.addEventListener(
--- a/src/app/index.tsx
+++ b/src/app/index.tsx
@@ -88,17 +88,14 @@ window.addEventListener(
   },
 );
 
-window.addEventListener(
-  'unhandledrejection',
-  // @ts-ignore
-  ({ reason }: PromiseRejectionEvent) => {
-    store.dispatch(
-      unhandledErrorThrown({
-        location: pick(location, 'pathname', 'search', 'hash'),
-        name: 'Unhandled Rejection',
-        message: String(reason),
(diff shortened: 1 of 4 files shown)


forabi:
Well, looks like the changes here do not actually fix the issue. It just does not happen all the time.

forabi:
As usual, I figure out solutions while trying to sleep :rofl: 

29571458_2036813029665398_5093930979069892740_n

I realized that the problem was not that the rejections were not being handled. They _were_. Otherwise, they would not have been included in the log batch. I mean, how could I not see this? :sweat_smile: 

The problem was that the `catch` handler was not being reached in `tryLoading`. We should add another catch handler in `componentWillUnmount`.

This `CancelationError` makes up 96% of all runtime errors according to the logs. We should see significant reduction in runtime errors starting today.

src/app/index.tsx
```diff
@@ -88,17 +88,14 @@ window.addEventListener(
...
-        location: pick(location, 'pathname', 'search', 'hash'),
-        name: 'Unhandled Rejection',
-        message: String(reason),
-        ...(isError(reason) ? reason : undefined),
-      }),
-    );
-  },
-);
+// @ts-ignore
+window.onunhandledrejection = ({ reason }: PromiseRejectionEvent) => {
```
wholesomedev, line 92: Is this still a relevant change?

wholesomedev (commented):
😂 nice finding, though!

I hope your brain will you sleep well now!

src/app/index.tsx
forabi: Probably not, but it seems that this is more widely supported by Promise polyfills.

Issue #514, opened 2018-05-30
We should have an automated way to detect issues like #512


forabi:
For JS errors that block rendering of the page, the logging epic and Google Analytics may not even load. We need to use some tool that can track visual regressions.

wholesomedev:
Looks like [BrowserStack](https://www.browserstack.com/screenshots) might give us this capability?

forabi:
Nice! I was thinking of some thing more automated though (like https://percy.io/). I'll take a look at a few solutions.

wholesomedev:
@forabi I looked at Percy. Based on my understanding, Percy doesn't support multiple browsers. It doesn't even let us choose a browser. It will render our DOM in a "modern browser" and if it sees pixel differences, it will notify us?

I don't see how this can be easily automated without a very intelligent and fine-tuned artificial intelligence.

The way I imagine this would work for us is kind of similar to how we used to do perf-reports. We would manually check a daily report of screenshots on different browsers.

forabi:
> It will render our DOM in a "modern browser" and if it sees pixel differences, it will notify us?

Percy was just one suggestion. We could probably find a similar service which allows choosing older browsers. If not, we could use BrowserStack with something like [pixelmatch](https://github.com/mapbox/pixelmatch) to diff screenshots on the same browser.

forabi:
I've looked more into this and I think it would be complicated to do this on CI because it would need a staging environment. I think the screenshot and diff with `track-performance` appraoch would work fine for us.

wholesomedev:
👍

forabi:
I'm going to re-open this until we add a panel to Splunk dashboard to report the screenshots diffs.

wholesomedev:
Yeah, I was wondering where we were going to post these images. Looks like Splunk allows [adding images](https://answers.splunk.com/answers/136162/add-picture-to-dashboard.html) to dashboards. Are we going to do something like that?

forabi:
That worked! (although we should maybe consider cropping the image :smile:)

![screenshot from 2018-06-13 12-55-24](https://user-images.githubusercontent.com/1012761/41344232-4271193a-6f09-11e8-886c-04c662829472.png)

forabi:
I improved this panel by adding overflow on the image and allowing the screenshot to be loaded dynamically instead of being hardcoded. This allows selecting the page URL:

[...]

PR #520, opened 2018-06-01
Revert "Use p-memoize instead of custom implementations (#516)"

forabi:
This reverts commit 6a44685b841990f7bc2c55604d4ab474441f1904.

I was using [Browserling](https://www.browserling.com/browse/win/7/chrome/41/https%3A%2F%2Fhollowverse.com%2FTom_Hanks) to test the website on Chrome 41 and it looks like this broke it again.

Debugging revealed that `p-memoize` uses ES2015+ syntax and it does not get compiled with Babel because `/node_modules/` are ignored.

I tried to include `node_modules` in `babel-loader` configuration but got a ton of build errors.

This makes #514 even more important.


wholesomedev (commented):
LGTM

>This makes #514 even more important.

Yep.

Issue #508, opened 2018-05-29
How should we decide on an error threshold?


wholesomedev:
I'm wondering if we can come up with a formula for calculating an acceptable error threshold. For example, with `PAGE_LOAD_FAILED`, we can say as long `PAGE_LOAD_FAILED` is lower than X% of total page views, we can tolerate it.

But what about run-time errors? Can we come up with a similar formula for that? The problem with run-time errors is that it's hard to tell what their impact is just by looking at the logs. One run-time error may prevent a button from working, another error could freeze the whole app, and another error may not have any user-perceptible impact.

Looking at our logs, it seems to me that it might be possible to shoot for for 0% tolerance level for run-time errors.

Alternatively, maybe we can define our tolerance for runtime errors based on a formula such as "total number of errors"/"total number of unique users". It's not a very meaningful formula, but it's the most meaningful one I could think of given the variability of runtime errors.

What are your thoughts on the tolerance levels we should shoot for for failed page views and runtime errors?

forabi:
> The problem with run-time errors is that it's hard to tell what their impact is just by looking at the logs.

This :100:. Just looking at the different errors I see too many with the message `[object Object]` on Chrome 66. I'm using Chrome 66 but I don't get that error. There are just too many variables in play. Maybe some users are using an extension which is throwing an error? We would not be able to know.

> Looking at our logs, it seems to me that it might be possible to shoot for for 0% tolerance level for run-time errors.

Interesting. I'd say it would be pretty hard to get to 0%. But I don't think ~9% (which is what we are currently at) is acceptable either. Something between 3-5% might be more realistic.

I think we need to start parsing stack traces to inform the decision here, although it seems that some of the errors that are logged do not have stack traces unfortunately. They could be non-`Error` objects being thrown.

With regard to `PAGE_LOAD_FAILED` events, these are mostly network errors according to the logs. We currently have around 2% of pages failing to load. That's OK I guess but I'd like to see it decreasing if possible. We currently show a Retry button so even if some page fails to load due to network errors, users might still be viewing it after reload. Maybe we could track `PAGE_LOAD_FAILED` vs `PAGE_LOAD_SUCCEEDED` for the same page in the same session to confirm that?

wholesomedev:
>Interesting. I'd say it would be pretty hard to get to 0%. But I don't think ~9% (which is what we are currently at) is acceptable either. Something between 3-5% might be more realistic.
>
[...]

PR #521, opened 2018-06-10
Check whether browser is supported in log events

forabi:
See https://github.com/hollowverse/hollowverse/issues/508#issuecomment-395986778


wholesomedev (commented):
LGTM!

PR #420, opened 2018-04-19
Enable GraphQL intellisense, TSLint rules that require type checking, ESLint for TS files

forabi:
* Enable descriptions, auto-completion, go to definition and linting for GraphQL queries. Powered by the already recommended "GraphQL for VS Code" extension.

  Note: requires Watchman, which is a native dependency (https://facebook.github.io/watchman)

  screenshot from 2018-04-19 01-12-58

* Enable TSLint as a language server plugin for TypeScript: which means rules that require information about the types of objects work in VS Code now!

  Note: requires experimental version of TSLint extension:
  
  
  screenshot from 2018-04-19 01-25-29
  
  screenshot from 2018-04-19 01-25-00

* Enable ESLint for TypeScript files!

  screenshot from 2018-04-19 06-58-47

PR #528, opened 2018-06-15
Enable GraphQL intellisense

forabi:
This enables the same features in #420 but with a new, more reliable extension (the other ESLint and TSLint features in that PR are already included in master).

Closes #420.


wholesomedev (commented):
LGTM!

PR #535, opened 2018-06-23
Implement app menu

forabi:
* [x] Keyboard accessible
  * [x] Allow focus rings with keyboard navigation, hide for mouse navigation
* [x] Follow best practices outline here: https://www.w3.org/WAI/tutorials/menus/
* [x] Prevent page scrolling when menu is shown?
* [x] Animations
* [x] Semantic markup
* [x] Validate HTML
  * [x] Is it okay to have an empty `<nav>` or menu?
* [x] Test on Safari for iOS
  * [ ] ~Workaround nav bar expanding on menu visible (maybe stop propagation of scroll events instead of hiding overflowing content?)~
* [x] Test on Chrome for Android
* [x] Test on a screen reader
* [ ] ~Connect with API query for `viewer`~
* [x] Hide for now until authentication is implemented


wholesomedev:
@forabi I thought about this, and I'm really not sure what to do about it. If we can stop scrolling in all browsers but iOS, I'm okay with that. Or if it's too difficult to stop scrolling for now, we can go with a degraded implementation and create an issue to track this.

forabi:
@wholesomedev I ended up using CSS `touch-action: none;` which achieves the desired behavior on browsers that support `none` for this property. Safari does not, so scrolling will still be possible. The upside for this CSS-based solution is that it won't cause the Safari navbar to expand when the menu is opened. So this is a best-effort solution for now.

wholesomedev:
Sounds good to me 👍

forabi:
@wholesomedev The changes in this PR are available for testing on https://hollowverse.com/Tom_Hanks?branch=internal

Feel free to test and provide feedback!

The menu is fully keyboard-accessible, and I've also tested it on Android and Safari screen readers. I've changed the animation from the one in design to a scaling animation. I just felt like it's snappier and more "appy".

src/app/components/App/App.module.scss
```diff
@@ -7,6 +7,11 @@
     scroll-behavior: smooth;
   }
 
+  .no-scroll {
+    // stylelint-disable-next-line plugin/no-unsupported-browser-features
+    touch-action: none;
```
wholesomedev: What about in desktop browsers, can we disallow scrolling while the menu is open in desktop browsers? The ability to scroll while the menu is open is causing some flickering behavior in macOS Chrome, especially at the moment when the our top bar animates from logo to search box.

src/app/components/AppMenu/AppMenu.module.scss
```diff
@@ -0,0 +1,99 @@
...
+      transition-duration: var(--timeout);
+      transition-timing-function: ease-in-out;
+    }
+    // #endregion Transitions
+
+    .body {
+      position: fixed;
+
+      // The actual position for the menu body will be set dynamically
+      // vis JS. See the `getMenuStyle` prop on the menu component.
```
wholesomedev, line 47: 🤯

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -0,0 +1,145 @@
...
+                {Separator}
+                <MenuItemWithLink size="small" to="/privacy-policy">
+                  Privacy Policy
+                </MenuItemWithLink>
+                <MenuItemWithChild
+                  size="small"
+                  factory="div"
+                  isClickable={false}
+                >
+                  &copy; 2018 Hollowverse
```
wholesomedev: Nit: since none of what we don't is copyrighted, I don't think we should have this line.

src/app/components/AppMenu/MenuItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+    className={cc([
+      classes.root,
+      className,
+      { [classes.small]: size === 'small' },
+    ])}
+    // Override role="menuitem" set by react-aria-menubutton because
+    // it causes Android TalkBalk to read the menu item and its children
+    // each separately. We will instead set `role="menuitem"` on the child
+    // below and make screen reader skip the parent entirely.
+    role="presentation"
```
wholesomedev: 💙

src/app/pages/SearchResults/ResultsList.tsx
```diff
@@ -26,7 +26,11 @@ export const ResultsList = (props: ResultsListProps) => {
     'isLoading' in props ? resultsListDummyData : props;
 
   return (
-    <ol className={classes.root} aria-hidden={isLoading}>
+    <ol
+      aria-label="Search Results"
+      className={classes.root}
+      aria-hidden={isLoading}
```
wholesomedev, line 32: Will the screen reader alert the user when the element becomes visible? Should something like `aria-alert` be used when content finishes loading and is now available to be consumed by screen reader?

wholesomedev (approved):
Wow, this is some next level stuff! Amazing attention to detail and accessibility. Awesome work!

src/app/components/App/App.module.scss
forabi: The flickering happens for me on Chrome and Firefox too. I think it has something to do with `z-index`es. If we set `overflow: hidden` on the body, it will:

* make Safari expand the navbar
* interfere with `scroll-behavior: smooth;`: when scrolling is restored, the browser will slowly scroll down to the previous scrolling position. Not a great experience.

I'll see what we can do.

src/app/components/AppMenu/AppMenu.tsx
forabi: Removed

src/app/pages/SearchResults/ResultsList.tsx
forabi: I was not actually sure about that. It occurred to me that we should use `aria-alert` or something similar, but I haven't learned much about it to do it in this PR.

forabi:
Okay, I've disabled scrolling on desktop as well using `overflow: hidden;` and excluded iOS devices. I've also fixed the issue of flickering when the navbar transforms into the search box (it had to do with stacking contexts and `z-index`).

One minor issue is that when `overflow: hidden;` is set, the content of the page shifts because the space that was occupied by the scrollbar becomes available for the document. I think the `no-scroll` library had a fix for this but in my testing it was not reliable. I'm going to check it again later.

I'm going to hide and disable the menu button for now. We can continue improving the menu in future PRs.

forabi:
By the way, I finally found the last missing piece of a great CSS development experience! SCSS IntelliSense extension for VS Code with support for variable auto-completion and go to definition!

screenshot from 2018-07-01 13-52-49

I've added the extension to the list of recommended extensions.

wholesomedev:
>I'm going to hide and disable the menu button for now. We can continue improving the menu in future PRs.

👍 

>By the way, I finally found the last missing piece of a great CSS development experience! SCSS IntelliSense extension for VS Code with support for variable auto-completion and go to definition!

Oooo nice! RIP CSS-in-JS 😜

PR #538, opened 2018-07-02
Implement user authentication

forabi:
* [x] Write tests
* [x] Figure out what's causing the build to break in Chrome 41


forabi:
It seems to me that the best way to avoid hitting FB rate limits is to not use the FB access token as our own API authentication token directly, and instead exchange the FB access token for something like a JSON web token that we generate on the API side and store in the database.

forabi:
@wholesomedev I found this gem in the [FB API rate limit doc page](https://developers.facebook.com/docs/graph-api/advanced/rate-limiting):

> The total number of calls your app can make per hour is 200 times the number of users.

So I guess we are fine using FB access tokens directly because the limit will increase as the number of user increases.

forabi:
@wholesomedev Okay, so here is an initial version with authentication and auto sign-up implemented https://hollowverse.com/George_Clooney?branch=internal

Feel free to provide feedback!

forabi:
@wholesomedev Note: there's currently a bug which causes the menu to not get updated after the account is created, so you may need to reload the page to see your profile in the menu.

wholesomedev:
@forabi This looks good!

Interestingly, I was logged in immediately upon opening https://hollowverse.com/George_Clooney?branch=internal I didn't experience the bug that you described.

Comments:

* I think we need to show some kind of feedback when a user logs in or out. Right now after I click, I don't get confirmation that the logout or login action succeeded.
* On COMB, navigating to home, contact, or other pages through the menu and then swiping back, the menu is initially open for a split second then disappears. Can we prevent that?
* An issue that's more related to the menu itself: the X button is on the same row as the Home button in the logged out state. That can be confusing. Should we move the X button to be closer to where the original menu open button is, that way it's more consistent? (Or maybe we shouldn't spend time working on this since we're moving to Material Design and a lot of this might change then anyway)

forabi:
> I think we need to show some kind of feedback when a user logs in or out. Right now after I click, I don't get confirmation that the logout or login action succeeded.

I think this is a perfect use case for the [Snackbar](https://material-ui.com/demos/snackbars/) component from Material Design.

forabi:
The other two issues you pointed out would probably be fixed when we use `material-ui`. I'm totally okay with replacing the menu and toggle components now with their `material-ui` counterparts and also adding the snackbar. However, both #539 and this PR are being blocked on the bug that's causing JS to break on Chrome 41. I'm going to look into this before we merge any pending PRs.

forabi:
So I checked the internal branch again with changes from this PR and it seems to work fine on Chrome 41 :thinking:. That's really weird.

forabi:
Okay, so now we have two branches, `hollowverse.com/Tom_Hanks?branch=night-mode` and `hollowverse.com/Tom_Hanks?branch=internal` and only the `night-mode` branch seems to be broken on Chrome 41. So it's safe to merge this when it's ready. We might as well start incorporating Material UI in this PR (e.g. add snackbar and replace the custom toggle implementation).

wholesomedev:
@forabi This all sounds good. But is there any chance that work on `hollowverse/design` might produce different UX choices? In that case, we might have to make alterations or rework this (I'm okay with making alterations or reworking this. I just thought I'd point out this possibility).

wholesomedev:
CC @chasepoirier

forabi:
@wholesomedev @chasepoirier I'm happy to wait for UX in design before migrating to Material Design here. I can write tests in the meantime.

wholesomedev:
@forabi By the way, in Light Mode, the container where the article is (the card) has higher contrast ratio with the text, whereas the general background of the page has lower contrast ratio. In Dark Mode, this is not the case. In Dark Mode, the general background has the higher contrast ratio with the text, whereas the article container (the card) has lower contrast ratio with the text.

I wonder if things would look better if we make the card darker than the general background in Dark Mode?

forabi:
@wholesomedev I've deployed a new version to `internal` with snackbars and dialogs from `material-ui`. The auth implementation is complete and tested. Visual changes are still needed though. Please feel free to test and report any issues.

forabi:
I'm going to merge this but keep the menu hidden for now. We can show it again after we have implemented quote submission.

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -25,59 +37,264 @@ const Separator = (
...
 
-export class AppMenu extends React.PureComponent<Props> {
+const messageForAuthState: Partial<Record<AuthState['state'], string>> = {
+  initializing: 'Checking login...',
+  loggingIn: 'Checking login...',
+  loggingOut: 'Logging out...',
+  loggedIn: 'Log out',
+  loggedOut: 'Log in with Facebook',
+  error: 'Log in with Facebook',
+};
```
wholesomedev, line 72: This is great! I just wonder if we should show the logging in/out statuses outside of the menu as well? I don't think it's a big deal though.

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -25,59 +37,264 @@ const Separator = (
...
+    if (authState.state !== 'loggedIn') {
       return undefined;
     }
 
+    const { viewer } = authState;
+
     return (
       <MenuItemWithChild
-        aria-label={`Signed in as ${user.name}`}
+        aria-label={`Signed in as ${viewer.name}`}
```
wholesomedev, line 152: Is `viewer` a GraphQL concept or a more modern name? I'm used to this object being called `user`

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -25,59 +37,264 @@ const Separator = (
...
+
+    return (
+      <Dialog
+        aria-labelledby="login-failed-dialog-title"
+        role="alertdialog"
+        onClose={this.toggleLoginFailedDialog}
+        open={this.state.isLoginFailedDialogShown}
+      >
+        <DialogTitle id="login-failed-dialog-title">
+          {titleForErrorCode[code] || titleForErrorCode.UNKNOWN_ERROR}
```
wholesomedev, line 207: This `||` branch here and below are unnecessary because you're defaulting `code` to `UNKNOWN_ERROR`

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -25,59 +37,264 @@ const Separator = (
...
+      requestLogin(undefined);
+    } else if (authState.state === 'error') {
+      this.toggleLoginFailedDialog();
+    }
+  };
+
+  renderLoginStateChangeSnackbar() {
+    const { authState } = this.props;
+
+    let message: React.ReactElement<any> | undefined;
```
wholesomedev, line 242: Do you not like `null` anymore? 😝

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -25,59 +37,264 @@ const Separator = (
...
+    } else if (authState.state === 'loggedOut') {
+      message = <span>Logged out</span>;
+    }
+
+    if (!message) {
+      return undefined;
+    }
+
+    return (
+      <Snackbar
```
wholesomedev, line 258: Later on, we should probably create our own lightweight wrapper around `Snackbar` to standardize things like `anchorOrigin` and `autoHideDuration`

src/app/store/features/auth/epic.ts
```diff
@@ -0,0 +1,134 @@
...
+  CreateUserMutation,
+  CreateUserMutationVariables,
+} from 'api/types';
+import { GraphQLClient } from 'graphql-request';
+import { merge, of as observableOf, defer, Observable } from 'rxjs';
+import { getAccessToken } from './reducer';
+
+import createUserMutation from './createUserMutation.graphql';
+import viewerQuery from './viewerQuery.graphql';
+import { setFbSdkAuthState, facebookAuthResponseChanged } from './actions';
```
wholesomedev, line 27: I don't think the logic that you use to break up the import statements is clear from an outsider's point of view. Do you think it might be good to follow a simpler logic such as:

```
// `node_module` dependencies

// Imports that use `baseUrl` (optionally alphabetically ordered)

// Relative imports
```

I'm just bikeshedding here. What you have is acceptable for now.

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -100,25 +317,7 @@ export class AppMenu extends React.PureComponent<Props> {
...
-                        color="var(--facebook-blue)"
-                        size={24}
-                        {...facebookIcon}
-                      />
-                    )
-                  }
-                >
-                  {user ? 'Log out' : 'Log in with Facebook'}
-                </MenuItemWithButton>
+                {this.renderLoginButton()}
```
wholesomedev, line 320: When will GitHub fix this syntax highlighting issue?! 🤦‍♀️

src/app/store/features/auth/epic.ts
```diff
@@ -0,0 +1,134 @@
...
+    method: 'POST',
+    cache: 'no-store',
+    headers: {
+      Authorization: fbAccessToken ? `Bearer ${fbAccessToken}` : '',
+    },
+  });
+
+  let viewerResult = await apiClient.request<ViewerQuery>(viewerQuery);
+
+  // If an access token is provider but not matching user is found,
```
wholesomedev, line 42: A couple typos

src/app/typings/fbSdk.d.ts
```diff
@@ -1,46 +1,113 @@
...
-
-    XFBML: {
-      parse(node?: Node, callback?: () => void): void;
-    };
-
-    getLoginStatus(
-      callback: (response: AuthResponse) => any,
-      cache?: boolean,
+/* tslint:disable */
+const FB: FB;
```
wholesomedev, line 2: Where does the `FB` after the colon come from?

src/app/typings/fbSdk.d.ts
```diff
@@ -1,46 +1,113 @@
...
+     * This indicates that all of the social plugins on the page have been loaded.
+     */
+    'xfbml.render': undefined;
+  };
+
+  type EventType = keyof EventTypeToResponse;
+
+  type Callback<T> = (response: T) => void;
+
+  type EventResponse<T extends EventType> = EventTypeToResponse[T];
```
wholesomedev, line 112: Are we going to contribute this to open source?

wholesomedev (commented):
Great work!!

src/app/components/AppMenu/AppMenu.tsx
forabi: It's actually a Relay concept and I think it conveys the meaning better than `user` which does not necessarily mean "the user who is doing the query".

src/app/components/AppMenu/AppMenu.tsx
forabi: Tell that to TypeScript 😂. Any prop access on `Partial<Record<K, V>>` will be of type `V | undefined` regardless of whether the key/value pair actually exists or not, so we'll have to provide a fallback value in places that don't accept `undefined`.

src/app/store/features/auth/epic.ts
forabi: These could be auto imports added by VS Code while typing and I forgot to re-organize them. I totally agree with the order you suggested but I just wish it was automated. I actually tried using the Organize Import feature in VS Code but it produced weird results because it doesn't understand `baseUrl` imports so they are handled like external packages and mixed with `node_modules` imports.

src/app/typings/fbSdk.d.ts
forabi: TS allows one name to be used for namespaces, variables and types. The second `FB` is used in a type annotation position so it refers to the type, which is defined below.

src/app/components/AppMenu/AppMenu.tsx
forabi: The defaults could be configurable via Material UI theme (not sure though).

src/app/typings/fbSdk.d.ts
forabi: I'd like to!

src/app/store/features/auth/epic.ts
forabi: Fixed in #548

PR #539, opened 2018-07-06
Add night mode

forabi:
(WIP, needs rebasing)

To do:
* [x] ~Figure out how to apply illumination effects with variable colors~ (not needed any more because we replaced illuminations with pulse effect)
* [x] Hide toggle for browsers that do not support CSS variables
* [x] Persist selected theme across page reloads
* [x] Log night mode clicks
* [ ] Test how changes look on Chrome 41 (which does not support CSS Custom Properties)
* [x] Figure out how to handle Facebook comments
* [x] Test color contrast


forabi:
I tried to style the photo for NPs missing photos by using an inline SVG instead of an `` tag but for some reason the photo didn't show in Chrome. It only worked in Firefox.

forabi:
I've tweaked the background and text colors for the night mode and tested the changes with [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd), [Color Contrast Analyzer](https://chrome.google.com/webstore/search/color%20contrast%20analyzer) and Lighthouse and they all seem to be happy with these values :smile:.

Note: [This page](https://dequeuniversity.com/rules/axe/2.2/color-contrast) also has a manual color contrast analyzer tool.

forabi:
I found a possible workaround for the FB comments widget -- instead of trying to meticulously find which elements inside the iframe to target (which could be changed by Facebook any time), I used the CSS `filter` property on the iframe, with the value: `grayscale(100%) invert(1);`

screenshot from 2018-07-10 21-17-50

One caveat with this approach: the images of commenters are also inverted.

Anyway, I think this is a much cleaner approach! What do you think, @wholesomedev?

I also found that `filter` could also work for the NP image placeholder, although not as well as FB comments (I used the value `invert(1) contrast(63%)`):

screenshot from 2018-07-10 21-23-55

forabi:
Since integrating Material UI breaks the website on Chrome 41 and will require more work than expected, and since we already have a toggle component implemented in #543, I'm going to remove `material-ui` packages from here and use the toggle from #543 for this PR. We can look into using `material-ui` later.

forabi:
Hmm, looks like it's not `material-ui` that's causing the website to break on Chrome 41. It's something we've added in recent PRs. The current master and beta are working fine on Chrome 41, but this PR is broken on the internal branch. It could be a dependency that we've added but not used until now, or it could be a minifier bug!

package.json
```diff
@@ -142,6 +142,7 @@
     "helmet": "^3.9.0",
     "history": "^4.6.3",
     "http-proxy-middleware": "^0.18.0",
+    "idb-keyval": "^3.0.5",
```
forabi: Looks like this is the dependency that's causing the `SyntaxError` on Chrome 41. The source maps were not useful at all. They were pointing to a completely different file.

wholesomedev:
@forabi By the way, in Light Mode, the container where the article is (the card) has higher contrast ratio with the text, whereas the general background of the page has lower contrast ratio. In Dark Mode, this is not the case. In Dark Mode, the general background has the higher contrast ratio with the text, whereas the article container (the card) has lower contrast ratio with the text.

I wonder if things would look better if we make the card darker than the general background in Dark Mode?

forabi:
Well, the good news is the website now works on Chrome 41... the bad news is Chrome 41 does not support CSS Custom Properties:

screenshot from 2018-07-13 21-35-26

A PostCSS plugin that we used to use supposedly inlines the values of CSS variables such that a fallback is used by browsers that do not support CSS variables. But it never actually worked because it's not smart enough to figure out that the variables used in some CSS file are actually defined in another file.

forabi:
This is mostly done, but a few issues remain unsolved:

* Chrome 41 does not support CSS Variables. This could be solved by the switch to Material UI which offers a JS-based styling system. We could define our night mode variant as a Material UI theme.
* The way the illumination effect is implemented does not work well with dark colors. The gradient just does not blend well with the background. I found another way to implement a similar effect but the best way I found involves animating the `top` and `left` properties of a `background-image` which is just incredibly inefficient compared to `transform` (I know because I tested both properties before in a different project). It also depends on browser support for `-webkit-background-clip: text` which is not standardized.

wholesomedev:
* I'm okay with doing this through Material UI, but that will require that we implement Material UI for all components that need this styling. Doing Material UI in a big chunk like this wasn't our original plan, but I'm okay with it in this case
* While our illumination effect is really cool, its implementation is not the most declarative (to put it mildly). If we can come up with a simpler animation that communicates the loading state and works with Night Mode, that would be a good middle ground in my opinion.

forabi:
Closing, superseded by #563

PR #543, opened 2018-07-07
Add toggle and dialog components

forabi:
I'm basically cherry-picking irrelevant commits from #539 and #538 so that they are easier to review.

### DIFF
--- a/src/app/components/Dialog/Dialog.module.scss
+++ b/src/app/components/Dialog/Dialog.module.scss
@@ -1,0 +1,26 @@
+@import 'variables.scss';
+
+.root {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  position: absolute;
+  pointer-events: none;
+
+  .child {
+    pointer-events: auto;
+    padding: $spacing-25px;
+    width: 75%;
+    min-width: 300px;
+    max-width: 400px;
+    min-height: 100px;
+
+    p {
+      margin-top: 15px;
+    }
+  }
+}
--- a/src/app/components/Toggle/Toggle.module.scss
+++ b/src/app/components/Toggle/Toggle.module.scss
@@ -2,72 +2,69 @@
 
 .root {
   display: flex;
-  align-items: stretch;
-  height: 100%;
-
-  --toggle-height: 1.2em;
-  --toggle-width: calc(var(--toggle-height) * 2.1);
-  --knob-size: calc(var(--toggle-height) - 2px);
-
-  input {
-    position: absolute;
-    opacity: 0;
-    width: 0;
-    height: 0;
-  }
-
-  .label {
+  button {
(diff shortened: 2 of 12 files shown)


src/app/components/Dialog/Dialog.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  .child {
+    pointer-events: auto;
+    padding: $spacing-25px;
+    width: 75%;
+    min-width: 300px;
+    max-width: 400px;
+    min-height: 100px;
+
+    p {
+      margin-top: 15px;
```
wholesomedev, line 23: Why not use `$spacing-15px` here?  Can we lint against literal pixel values?

src/app/components/Toggle/Toggle.module.scss
```diff
@@ -0,0 +1,83 @@
...
+
+  .label {
+    flex-grow: 1;
+    cursor: inherit;
+    @include truncate-flex-child();
+  }
+
+  --toggle-height: 1.15em;
+  --toggle-width: calc(var(--toggle-height) * 2.1);
+  --knob-size: calc(var(--toggle-height) - 0.1em);
```
wholesomedev, line 27: Why do these need to be defined as variables?

src/app/typings/react-aria-modal.d.ts
```diff
@@ -0,0 +1,151 @@
...
+     * have the function that closes the modal right there, written by you, at your disposal.
+     */
+    onExit(): any;
+  };
+  class AriaModal extends React.PureComponent<AriaModalProps, object> {
+    static renderTo(node: HTMLElement | string): void;
+  }
+
+  export default AriaModal;
+}
```
wholesomedev, line 151: Are we going to contribute this to other open source eventually?

wholesomedev (commented):
A couple of questions, otherwise LGTM!

src/app/components/Dialog/Dialog.module.scss
forabi: That's an oversight. We can force use of variables in specific CSS properties using [`stylelint-declaration-strict-value`](https://github.com/AndyOGo/stylelint-declaration-strict-value) or [`stylelint-declaration-use-variable`](https://github.com/sh-waqar/stylelint-declaration-use-variable). We can include one of these in hollowverse/config#93

src/app/components/Toggle/Toggle.module.scss
forabi: They don't. It's just convenient to be able to calculate some values based on another variable's value and be able to test the changes in the browser directly, instead of using build-time Sass variables which would require re-compilation every time.

src/app/typings/react-aria-modal.d.ts
forabi: Sure :smile: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27228

PR #545, opened 2018-07-09
Set up material-ui

forabi:
(pending implementation of FAB in hollowverse/design)


forabi:
Gonna close this as it's already included in #539

Issue #547, opened 2018-07-16
Migrate tests from Enzyme to react-testing-library


forabi:
[`react-testing-library`](https://github.com/kentcdodds/react-testing-library) has a nicer API and provides some utilities that are not available in Enzyme like `waitForElement`.

wholesomedev:
Should this be closed?

forabi:
I thought this would be automatically closed when the latest PR is merged.

Issue #541, opened 2018-07-07
Menu items cannot be selected with the Enter key


forabi:
Relevant issue: https://github.com/davidtheclark/react-aria-menubutton/issues/85

wholesomedev:
Should this be closed?

PR #548, opened 2018-07-20
Refactor tests, replace some components with material-ui counterparts

### LINKED ISSUE #540: Menu's fixed position does not update when the browser window is resized

forabi:
* [x] Make menu auto close on item selection and navigation
  * [x] Selection
  * [x] Navigation
* [x] Test accessibility

Fixes #540 #541 #547


src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -148,42 +128,41 @@ export class AppMenu extends React.PureComponent<Props, State> {
...
+        id="login-button"
+        button
         className={cc([
           {
             [classes.facebook]: state === 'loggedOut' || state === 'error',
           },
         ])}
-        type="button"
-        onClick={this.handleLoginClick}
+        onClick={callAll(this.handleLoginClick, this.handleClose)}
```
forabi: I really find it weird that we have to manually close the menu on item selection :thinking:

wholesomedev: @forabi Maybe that's because it's a controlled component that doesn't keep any state?

wholesomedev: ...or maybe these are Material Design guidelines?

forabi: @wholesomedev But the menu is still provided an `onClose` handler. So there should be no problem calling it whenever an item inside it is clicked. Right? :man_shrugging:

forabi: > ...or maybe these are Material Design guidelines?

I don't think so.

forabi: Another minor issue I found: we can't set a `max-width` on the menu and have the text be truncated:

localhost_8080_ 1

(unlike our custom implementation :sunglasses:)

hollowverse com_tom_hanks_branch internal

forabi:
I'm trying to come up with a good, maintainable way to have menus close automatically on navigation. It's not handled automatically, and I don't like having the menu know about location changes. Can you think of a better way? @wholesomedev

src/app/components/AppMenu/AppMenu.tsx
wholesomedev: I feel like we should at least log an issue for this to be addressed eventually? 😄

wholesomedev:
@forabi how about a wrapper menu that adds this behavior to material-ui menu?

forabi:
> how about a wrapper menu that adds this behavior to material-ui menu?

I kind of hate having to do this because it's still easy to import the wrong menu component and forget about this wrapper. The less wrappers we have, the better. But it looks like we'll have to do this.

wholesomedev:
I wonder if we can disallow certain imports for such cases, like by linting against them?

forabi:
> I wonder if we can disallow certain imports for such cases, like by linting against them?

[ESLint rules](https://eslint.org/docs/rules/no-restricted-imports) to the rescue!

src/app/__tests__/authInit.ts
```diff
@@ -25,39 +22,28 @@ describe('authentication', () => {
...
-      .find('button')
-      .filterWhere(el => !!el.text().match(/log in/i))
-      .simulate('click');
-
-    expect(
-      document.querySelector('[role="alertdialog"]')!.textContent,
-    ).toContain('tracking protection');
+    fireEvent.click(await context.toggleAppMenu().getLoginButton());
+
+    expect(document.querySelector('[role="alertdialog"]')).toHaveTextContent(
```
wholesomedev, line 45: Interesting. Usually, I would use Enzyme to find things in React.

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -284,56 +268,59 @@ export class AppMenu extends React.PureComponent<Props, State> {
...
-          {({ isOpen }: { isOpen: boolean }) => (
-            <CSSTransition
-              classNames={transitionClassNames}
-              timeout={transitionTimeoutMilliseconds}
-              in={isOpen}
-              mountOnEnter
-              unmountOnExit
+          <MenuIcon />
+        </IconButton>
+        {!!anchorElement && (
```
wholesomedev, line 295: Are we okay with this pattern now? 😜

src/app/components/AppMenu/AppMenu.tsx
```diff
@@ -148,42 +133,41 @@ export class AppMenu extends React.PureComponent<Props, State> {
...
-        isClickable={false}
-      >
-        <PersonPhoto
-          alt="Profile Photo"
-          className={classes.userAvatar}
-          src={viewer.photoUrl || undefined}
-        />
-        <div className={classes.userName}>{viewer.name}</div>
-      </MenuItemWithChild>
+      <InertMenuItem aria-label={`Signed in as ${viewer.name}`}>
```
wholesomedev, line 136: Cool name for the behavior. Had to look up the meaning of "Inert" lol

src/app/components/LocationAwareMenu/LocationAwareMenu.tsx
```diff
@@ -0,0 +1,28 @@
...
+      });
+    }
+
+    render() {
+      const { history, staticContext, match, location, ...rest } = this.props;
+
+      return <Menu {...rest} />;
+    }
+  },
+);
```
wholesomedev, line 28: I thought our implementation would be to call `onClose` for any `MenuItem` click

src/app/components/NavBar/NavBar.tsx
```diff
@@ -92,24 +55,22 @@ export const NavBar = class extends React.Component<
...
-                          className={cc([
-                            classes.button,
-                            { [classes.isHidden]: isHomePage },
-                          ])}
-                          to="/search"
+                        <IconButton
+                          aria-label="Search"
+                          className={cc([{ [classes.isHidden]: isHomePage }])}
+                          component={Link as any}
+                          {...{ to: '/search' } as any}
```
wholesomedev, line 70: Thanks TypeScript?

src/app/helpers/loadableDefaultOptions.tsx
```diff
@@ -1,11 +1,22 @@
...
   loading({ pastDelay }) {
     if (pastDelay) {
-      return <LoadingSpinner />;
+      return (
+        <div
+          style={{
+            width: '100%',
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
```
wholesomedev, line 14: Is it okay to have this in `style` attribute rather than in a stylesheet?

wholesomedev (commented):
A lot of good RED in this PR!

Nice use of Material UI, and other great stuff!

src/app/__tests__/authInit.ts
forabi: I thought about this a lot and I now believe it's better to test the DOM. Actually, `react-testing-library` encourages this practice. Even if we were to test the render container returned from `render` in test context, we would still need to check for some elements outside that container -- for example, Material-UI dialogs are always rendered outside the React app container for accessibility purposes.

src/app/helpers/loadableDefaultOptions.tsx
forabi: Well, since it's a loading spinner, I don't think it makes sense to have to load two files (JS, CSS) before rendering the spinner.

src/app/components/NavBar/NavBar.tsx
forabi: 🤷‍♂️. We should probably abstract this with a wrapper component

src/app/components/LocationAwareMenu/LocationAwareMenu.tsx
forabi: This is a different issue. We should probably also wrap `MenuItem` to call `onClose`.

PR #554, opened 2018-07-27
Continue to replace components with material-ui counterparts

forabi:
* [x] Improve spacing values
* [x] Increase body font size?
* [x] Customize search box
* [x] Fix NavBar sticky behavior
* [ ] ~Make background white?~
* [ ] ~Use paper for main NP image?~
* [x] Remove links underlines
* [x] Hide menu


forabi:
@wholesomedev Okay, so here we go! Please review the changes on https://hollowverse.com/Tom_Hanks?branch=materialize-more and let me know what we can improve!

I replaced the illumination effect with a "pulse" effect which I've copied from the Google News app. I think it'll work better with night mode.

wholesomedev:
@forabi I took a peek and I think it looks good! For some reason, the layout and interactions feel more cohesive.

I’ll take a closer look later tonight and leave some feedback on margins, sizes, etc. 

Awesome work! 👍

wholesomedev:
Feedback on new design:

* The background color in the current Hollowverse is darker than the new one. I think the darker color is better. 
* Are we going to remove our cover colors?!
* Can we add some elevation to the top navbar? At least same level of elevation as the Paper component?
* Can we reduce the height of the navbar?
* The tooltips on the Main Menu and Search are not very legible. Can we increase their size or remove them?
* Can we maintain the typography, font-sizes, colors, and spacing of the current Hollowverse?
* The "Sources" text and button moves up and down between being open and closed. The separator also disappears and appears between open and closed. Can we keep these aspects the same between the open and closed states?
* Should we put the Related People inside a Paper?
* The Main Menu is too narrow when the logged in user has a short name

forabi:
Great feedback @wholesomedev. I think the answer to these points depends on whether we are fully adopting the Material Design spec (and if so, which version) or just using `material-ui` as a UI component library.

> * The background color in the current Hollowverse is darker than the new one. I think the darker color is better.
> * Can we add some elevation to the top navbar? At least same level of elevation as the Paper component?

If you look at the tasks above, I was actually going to make the background even lighter 😅, and I've intentionally removed the elevation on the navbar. Both changes bring the design closer to the new material design spec, which is already used in the [Google News app](https://play.google.com/store/apps/details?id=com.google.android.apps.magazines&hl=en). The navbar blends with the background and only elevates on scroll, which is what I'm trying to mimic.

> Are we going to remove our cover colors?!

This again depends on what we are adopting (the spec or just the components) and if we are going to bring back the cover colors, we won't be able to blend the navbar with the background.

Another solution we could use is to make the navbar and the FAB use the colors extracted from the image. This is a very common Material Design pattern.

> Can we reduce the height of the navbar?

Again, depends on whether we are adopting the spec (and whether the spec allows for height customization)

> The tooltips on the Main Menu and Search are not very legible. Can we increase their size or remove them?
> Can we maintain the typography, font-sizes, colors, and spacing of the current Hollowverse?

These should be easy to customize.

> The "Sources" text and button moves up and down between being open and closed. The separator also disappears and appears between open and closed. Can we keep these aspects the same between the open and closed states?

This is the `ExpansionPanel` component (which, by the way, has been removed from the new material design spec) and I haven't customized it at all -- the separator is part of the component. I'll look into that.

> Should we put the Related People inside a Paper?

If we make the background white, it'll look much like Google News:

image

> The Main Menu is too narrow when the logged in user has a short name

This should be customizable.

Let me know what you think. I don't have strong preferences here, I'm just listing all the possible options that we have.

wholesomedev:
> Great feedback @wholesomedev. I think the answer to these points depends on whether we are fully adopting the Material Design spec (and if so, which version) or just using `material-ui` as a UI component library.

I don't personally have an opinion here.

>
> > * The background color in the current Hollowverse is darker than the new one. I think the darker color is better.
> > * Can we add some elevation to the top navbar? At least same level of elevation as the Paper component?
>
> If you look at the tasks above, I was actually going to make the background even lighter , and I've intentionally removed the elevation on the navbar. Both changes bring the design closer to the new material design spec, which is already used in the [Google News app](https://play.google.com/store/apps/details?id=com.google.android.apps.magazines&hl=en). The navbar blends with the background and only elevates on scroll, which is what I'm trying to mimic.

The Google News app on Android looks pretty. The iOS version isn't nearly as pretty 😄 

I think it's a good idea to copy a professionally designed app, and I would agree that in that case, it might not be wise to deviate too much from the design we're trying to mimic.

One thing I think we should keep in mind is that we're trying to come up with a design that looks good on both mobile and desktop. I find that the new design looks better on mobile than on desktop. For example, the taller navbar is causing the logo, menu, and search icons to be perceived as smaller, which doesn't look very good. 

Not sure how we can improve the situation of mobile vs desktop. We might just have to compromise the desktop experience.

> > Are we going to remove our cover colors?!
>
> This again depends on what we are adopting (the spec or just the components) and if we are going to bring back the cover colors, we won't be able to blend the navbar with the background.
>
> Another solution we could use is to make the navbar and the FAB use the colors extracted from the image. This is a very common Material Design pattern.

It's just that we spent a lot of time getting the cover image working and it looks pretty, but ultimately, I'm okay with any approach here.

> > Can we reduce the height of the navbar?
>
> Again, depends on whether we are adopting the spec (and whether the spec allows for height customization)

Only problem with the big navbar is that it's causing its content to be perceived as smaller, but again, no strong opinion here.

> > The tooltips on the Main Menu and Search are not very legible. Can we increase their size or remove them?
> > Can we maintain the typography, font-sizes, colors, and spacing of the current Hollowverse?
>
> These should be easy to customize.

Another enhancement we can do here is to add a delay before the tooltip is displayed.

> > Should we put the Related People inside a Paper?
>
> If we make the background white, it'll look much like Google News:
>
> image

👍 

> Let me know what you think. I don't have strong preferences here, I'm just listing all the possible options that we have.

I don't have a strong preference here either. Again, if we're going to copy Google News, I think it's better to be careful with deviations we make.

I personally prefer legible layouts where there are clear demarcations of the UI sections, but I noticed that in newer designs things blend together more.

forabi:
> One thing I think we should keep in mind is that we're trying to come up with a design that looks good on both mobile and desktop. I find that the new design looks better on mobile than on desktop. 

Couldn't agree more! All the updated Google apps focus primarily on mobile layouts. It looks like they just gave up on the tablet business 😄

forabi:
> The Google News app on Android looks pretty. The iOS version isn't nearly as pretty

By the way, you could be using the old Google News app. Here is [the new "AI-powered" one](https://itunes.apple.com/us/app/google-news/id459182288?mt=8). Yeah, Google is that fragmented.

wholesomedev:
Ah, yeah, I didn't think to look for Google's other news app that's also called Google News 😄 

I mean I get it if Google wants to allow different teams to have their take on the same problem, but can't they come up with distinct names?

Anyway, yeah, this app looks much better! But now that I look at it, I can see that we might need to create interactions for which there are no equivalents on the Google News app. In that case, we might have to mix and match designs from different apps!

forabi:
> Anyway, yeah, this app looks much better! But now that I look at it, I can see that we might need to create interactions for which there are no equivalents on the Google News app. In that case, we might have to mix and match designs from different apps!

We are a little bit unlucky to be in this awkward time when apps are transitioning from Material Design v1 to v2.

forabi:
* [x] Restored the darker background color
* [x] Restored the navbar elevation
* [x] Restored the cover background
* [ ] It's not possible to reduce the navbar height. The icon buttons don't shrink to fit inside the navbar. At least not in a straightforward way.
* [x] Increased the size of tooltips
* [x] Delayed tooltip appearance
* [x] Wrapped the Related People section in a Paper.

wholesomedev:
@forabi I think it looks good! What do you think?

>It's not possible to reduce the navbar height. The icon buttons don't shrink to fit inside the navbar. At least not in a straightforward way.

I actually thought we should reduce the navbar height without changing the icon sizes. I noticed that the icons on the new design appear smaller than the icons in the current design (even though they're the same size) because of the increased height of the navbar. If we can reduce the height of the navbar, then its content won't appear smaller anymore.

forabi:
I meant they flow outside the navbar, they do not move to be vertically
centered.

wholesomedev <notifications@github.com> schrieb am So., 29. Juli 2018, 8:13
nachm.:

wholesomedev:
@forabi I think in that case it's okay to leave them as they are.

forabi:
> Another enhancement we can do here is to add a delay before the tooltip is displayed.

Done.

> The Main Menu is too narrow when the logged in user has a short name

Fixed.

> Can we maintain the typography, font-sizes, colors, and spacing of the current Hollowverse?

> The "Sources" text and button moves up and down between being open and closed.

Looking into these!

> The separator also disappears and appears between open and closed.

Done.

forabi:
> I actually thought we should reduce the navbar height without changing the icon sizes. 

I've managed to do just that. Turns out I had to customize the `Toolbar` component inside the `AppBar`. It has a `dense` variant which makes its height 48px, identical to what we originally had.

forabi:
@wholesomedev I think the changes on this branch look good. Let me know what you think.

wholesomedev:
@forabi I think these improvements look good!

There might be still some improvements we can make to the text.

### Current:

image

### New:

image

I feel the text looks better and is more legible in the current one?

Some of the things that look different to me are line spacing, font size, and spacing between paragraphs and headings.

---

The summary in the current one is also better, in my opinion.

### Current:

image


### New:

image

forabi:
I'll update those with theme overrides if possible. Otherwise, I'll have to use inline styles. Using our custom stylesheets with MUI's injected ones is not working well due to CSS specificity rules.

forabi:
The font kerning/anti-aliasing looks different to me between the current and new designs. Do you think that's the case? I can't find any CSS property set by Material UI that would cause this :thinking:

forabi:
@wholesomedev I've tweaked the margins and font styles in the editorial summary to be closer to the current design. I've also cusomtized the search box in the navbar to match the current one. Let me know what you think!

wholesomedev:
@forabi I think this looks good!

I like that we've restored the original styles of the input field.

The new font sizes and margins also look good to me 👍

I think we can go live with this new design. It's pretty exciting that we've migrated our UI to Material-UI!

> The font kerning/anti-aliasing looks different to me between the current and new designs. Do you think that's the case? I can't find any CSS property set by Material UI that would cause this 🤔

I've had this issue before. It's very bizarre. I was able to eventually locate it's source by going over all the CSS properties and toggling them on and off. I still don't remember the exact property that was responsible for this 😄

forabi:
> It's pretty exciting that we've migrated our UI to Material-UI!

Well, not completely... a few more components can be rewritten to use Material UI and I'd like to get rid of most of our SCSS files in favor for MUI's styling solution (JSS) because they do not play well with MUI's injected stylesheets. However, we can merge this now and continue migrating styles in later PRs!

deploy.js
```diff
@@ -12,7 +12,7 @@ const { IS_PULL_REQUEST, BRANCH } = shelljs.env;
 
 const isPullRequest = IS_PULL_REQUEST !== 'false';
 
-const whitelistedBranches = ['master', 'beta', 'internal'];
+const whitelistedBranches = ['master', 'beta', 'internal', 'materialize-more'];
```
wholesomedev, line 15: We should probably remove `materialize-more`.

src/app/components/Input/Input.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const Input = withStyles(styles)(
+  class extends React.Component<Props> {
+    render() {
+      const { classes, ...rest } = this.props;
+
+      return <DefaultInput disableUnderline classes={classes} {...rest} />;
+    }
+  },
+);
```
wholesomedev, line 29: Are we going to be using this style for all of our input fields, or will we eventually want to separate regular input fields from search bar input field?

Anyway, just something to keep in mind for the future.

src/app/helpers/testHelpers.tsx
```diff
@@ -244,11 +256,18 @@ export const createTestTree = ({
...
-      <Provider store={store}>
-        <ConnectedRouter history={history}>
-          <App routesMap={routesMap} />
-        </ConnectedRouter>
-      </Provider>
+      {/*
+        We use a customied JSS provider here to produce predictable class names in tests
+        for snapshot tests
+        See: https://github.com/mui-org/material-ui/issues/9492#issuecomment-368205258
+        */}
```
wholesomedev, line 263: 👍

wholesomedev (commented):
LGTM, nice work!

PR #558, opened 2018-08-08
Use visibilitychange event to send logs

forabi:
This works more reliably than the `unload` event across desktop and mobile browsers.

See https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/

### DIFF
--- a/src/app/store/features/logging/epic.ts
+++ b/src/app/store/features/logging/epic.ts
@@ -123,17 +123,21 @@ export const loggingEpic: Epic<Action, StoreState, EpicDependencies> = (
     const flushOnPageVisibilityChange$ = merge(
       // See https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/
       observableFromEvent(document, 'visibilitychange').pipe(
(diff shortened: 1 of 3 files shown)


src/app/store/features/logging/epic.ts
```diff
@@ -120,13 +120,17 @@ export const loggingEpic: Epic<Action, StoreState, EpicDependencies> = (
...
+      // See https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/
+      observableFromEvent(document, 'visibilitychange').pipe(
+        filter(() => document.visibilityState === 'hidden'),
+      ),
       observableFromEvent(window, 'pagehide'),
-      observableFromEvent(window, 'unload'),
     );
 
-    const logOnUnload$ = loggableActions$.pipe(buffer(flushOnUnload$));
+    const logOnUnload$ = loggableActions$.pipe(
```
wholesomedev: Should we change “Unload” to something else in this variable name?

forabi: Updated

wholesomedev (commented):
LGTM!

PR #561, opened 2018-08-13
Move Sass styles to JSS

forabi:


forabi:
Whoa! I expected the development experience of CSS-in-JS to be better than Sass. But now that I've started migrating more components to JSS, I'm convinced that CSS-in-JS has no real benefits over CSS Modules + Sass _and_ it is an inferior development experience.

forabi:
I'm actually starting to wonder if migrating Sass styles to JSS is worth it. Other than some pain points with integration between the two styling solutions (which I think we could work around with CSS variables) and some issues with CSS specificity, I think the overall development and user experience of Sass is actually superior.

wholesomedev:
@forabi maybe I don't completely understand how the Sass-JSS combination looks like. But I feel that having a single styling technology in the codebase may be the better setup.

In my opinion, a codebase that embraces an inferior technology fully is usually better than a codebase that has a contest between an inferior and a superior technology.

forabi:
> But I feel that having a single styling technology in the codebase may be the better setup.

That's the reason I wanted to move Sass styles to JSS. But I just assumed the experience would be better because of all the JSS-in-CSS hype, but it actually turned out to be worse. 

For example:
* there's no syntax highlighting for JSS
* very basic validation of CSS property values (using TypeScript) compared to CSSTree Validator.
* nesting selectors is inconvenient
* type checking with TypeScript is very restrictive and can get in the way
* looks like it's not possible to use non-class selectors (e.g. `.root p`)
* inability to lint CSS using stylelint
* Linting against unused code works better in Sass using stylelint plugins. In JSS, we don't get a warning when we keep an used class name.

I'm not against migrating to JSS, but I just wanted to point out these issues. It's something to keep in mind. I'll continue working on this.

wholesomedev:
These are significant drawbacks. Thanks for pointing them out.

As you said, we can continue the migration and hopefully gain more clarity as we keep going.

.stylelintrc.json
```diff
@@ -1,16 +0,0 @@
...
-        "ignore": [
-          "css-featurequeries",
-          "outline",
-          "css-appearance",
-          "viewport-units"
-        ]
-      }
-    ]
-  }
-}
```
wholesomedev, line 16: Bye bye stylelint! 👋

wholesomedev (commented):
This LGTM! Thanks for doing this migration.

I'm a little bit concerned about the fact that all styles are literal values.

Things might look unprofessional if we intend for two different elements to have the same styles but then we write the styles where some units are off by a few numbers here and there.

The `$spacing-` and `$font-size-` variables were meant to help with this issue.

It would be good if we can introduce some reusability to the styles in a way that makes sense so that we're not rewriting them in slightly different configurations unintentionally in different places.

forabi:
> I'm a little bit concerned about the fact that all styles are literal values.

I'm aware of this, but I wanted to get things working first. In the next iteration, I'll try to figure out the best way to pass custom variables. Ideally, they should be added to the `theme` object which gets passed to the function that calls `createStyles`, but I'm not sure TypeScript will be happy with this. Another option is CSS Variables.

PR #562, opened 2018-08-17
Fix NP image size, navbar content on small screens

forabi:
screen shot 2018-08-17 at 09 56 00

screen shot 2018-08-17 at 10 06 33


wholesomedev (commented):
Good catch, LGTM!

PR #575, opened 2018-09-07
Switch to uppercase enum values for API types

forabi:
I've updated the data in the database to use uppercase types for `EditorialSummaryNodeType`. I've also invalidate the API cache to avoid any problems with outdated queries.


wholesomedev (commented):
LGTM

Issue #569, opened 2018-09-05
We are approaching the maximum Lambda code package size


forabi:
The zip file that gets uploaded to S3 and then deployed to Lambda for `hollowverse/hollowverse` is approaching 50 MB, which is [the maximum limit for AWS Lambda packages](https://docs.aws.amazon.com/lambda/latest/dg/limits.html).

```
Serverless: Uploading CloudFormation file to S3...
Serverless: Using S3 Transfer Acceleration Endpoint...
Serverless: Uploading artifacts...
Serverless: Uploading service .zip file to S3 (48.61 MB)...
```

The zip file includes the server-side components as well as the client side assets. The server-side code is not currently bundled with webpack and instead, all required `node_modules` are zipped as-is. We should probably bundle those with webpack as well so we could drop unused files.

PR #578, opened 2018-09-08
Bundle server code with webpack

forabi:
This splits the deployment package into 2 ZIP files (one for each function), each is around ~7MB. The package size for `serveLogEndpoint` could be reduced even further if there was an easy way to [make `serverless-webpack` respect the per-function `include` option of Serverless](https://github.com/serverless-heaven/serverless-webpack/issues/333?ts=2).

Fixes #569


serverless.yml
```diff
@@ -1,11 +1,22 @@
...
 
 custom:
+  scripts:
+    hooks:
+      'before:package:createDeploymentArtifacts': yarn run-s copy-defs app/build
+
+  webpack:
+    packager: 'yarn'
+    webpackConfig: ./src/webpack/webpack.config.server.js
+    includeModules: true
```
wholesomedev, line 18: If webpack builds a single JS bundle, that includes all the dependencies, why do you need to `includeModules`?

wholesomedev (commented):
LGTM!

serverless.yml
forabi: Because there are still some packages not bundled with webpack due to incompatibilities. Those are defined in the `externals` field of the webpack config.

PR #585, opened 2018-09-16
Default to dark theme if system-wide dark mode is enabled

forabi:
WebKit has recently added support for detecting whether system-wide dark mode is enabled via [the new `prefers-dark-interface` media query](https://mobile.twitter.com/Keithamus/status/1007557906378706944?s=17). This is only available on Safari on macOS Mojave currently. I don't have a macOS device to test this but it should not break anything.

### DIFF
--- a/src/app/store/createConfiguredStore.ts
+++ b/src/app/store/createConfiguredStore.ts
@@ -126,7 +126,7 @@ const defaultInitialState: StoreState = {
   fbSdkAuthState: {
     state: 'initializing',
   },
-  theme: 'light',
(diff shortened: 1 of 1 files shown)


forabi:
Alas, I got too excited! It looks like this is a private API and would not be exposed in WebKit releases. Still waiting for a serious standardization effort. Let's hope that [this  issue](https://github.com/w3c/csswg-drafts/issues/2735) sees some progress soon.

wholesomedev:
I +1'd it 🤞

